PR #30, opened 2025-04-17
Add teams functionality

PhiRho:
Add the concept of teams, and update users to be associated with them. Also adds the concept of team roles, which allows us to put AuthZ on different levels. Currently we allow team-admins to perform certain actions in a more limited capacity than system admins.

**Note** all the frontened javascript is written by cursor without much more than visual testing (and linter validation) so may not be to a good standard.


app/api/teams.py
```diff
@@ -0,0 +1,178 @@
...
+from app.db.models import DBTeam, DBUser
+from app.schemas.models import (
+    Team, TeamCreate, TeamUpdate,
+    TeamWithUsers
+)
+from app.api.auth import get_current_user_from_auth
+
+router = APIRouter()
+
+def check_admin(current_user: DBUser):
```
czue, line 16: one thing I like to do is turn this type of thing into a decorator so it doesn't clutter the internals of a view's business logic. though that does require some assumptions about how the user will appear in the view args, which I'm not sure are valid/easy in FastAPI.

app/api/teams.py
```diff
@@ -0,0 +1,178 @@
...
+    return True
+
+@router.post("", response_model=Team, status_code=status.HTTP_201_CREATED)
+@router.post("/", response_model=Team, status_code=status.HTTP_201_CREATED)
+async def register_team(
+    team: TeamCreate,
+    db: Session = Depends(get_db)
+):
+    """
+    Register a new team. This endpoint is publicly accessible.
```
czue, line 45: Was surprised this doesn't require auth, but probably just because i'm still ramping up on what checks happen where in the onboarding flow.

app/db/models.py
```diff
@@ -40,12 +41,31 @@ class DBUser(Base):
     hashed_password = Column(String)
     is_active = Column(Boolean, default=True)
     is_admin = Column(Boolean, default=False)
-    created_at = Column(DateTime, default=datetime.utcnow)
+    created_at = Column(DateTime(timezone=True), default=func.now())
+    role = Column(String, default="user")  # user, admin, key_creator, read_only
+    team_id = Column(Integer, ForeignKey("teams.id", name="fk_user_team"))
```
czue, line 46: users are never in multiple teams? if that ever changes this would require some reworking

app/db/models.py
```diff
@@ -40,12 +41,31 @@ class DBUser(Base):
...
+    updated_at = Column(DateTime(timezone=True), onupdate=func.now())
 
+    team = relationship("DBTeam", foreign_keys=[team_id])
```
czue, line 49: maybe my own ignorance, but why do you need a team and a team_id? i'd have thought they might be redundant

app/db/models.py
```diff
@@ -40,12 +41,31 @@ class DBUser(Base):
...
     api_tokens = relationship("DBAPIToken", back_populates="owner")
     audit_logs = relationship("DBAuditLog", back_populates="user")
 
+class DBTeam(Base):
+    __tablename__ = "teams"
+
+    id = Column(Integer, primary_key=True, index=True)
+    name = Column(String, index=True)
+    email = Column(String, unique=True, index=True)
+    phone = Column(String, nullable=True)
```
czue, line 60: nit: i find it a bit surprising that something called a team would have an email and phone number. maybe `admin_email` or similar? is there a reason this is separate from the admin user's contact info which you could get through an FK relationship?

app/db/models.py
```diff
@@ -59,15 +79,19 @@ class DBPrivateAIKey(Base):
...
     owner_id = Column(Integer, ForeignKey("users.id"))
     region_id = Column(Integer, ForeignKey("regions.id"))
-    created_at = Column(DateTime, default=datetime.utcnow)
+    created_at = Column(DateTime(timezone=True), default=func.now())
+    updated_at = Column(DateTime(timezone=True), onupdate=func.now())
+    team_id = Column(Integer, ForeignKey("teams.id"))
 
     owner = relationship("DBUser", back_populates="private_ai_keys")
     region = relationship("DBRegion", back_populates="private_ai_keys")
+    team = relationship("DBTeam", back_populates="private_ai_keys")
```
czue, line 88: in other projects I've found it useful to have a `BaseTeamModel` class that has all the fields needed to belong to a team. this makes it a bit more explicit when you want to handle things like permission checks in a generic/central way. May not be worth it for this project, though, especially if each one needs a unique `back_populates`...

app/api/teams.py
```diff
@@ -0,0 +1,178 @@
...
+    # Get users associated with the team
+    team_users = []
+    for user in db_team.users:
+        team_users.append({
+            "id": user.id,
+            "email": user.email,
+            "is_active": user.is_active,
+            "is_admin": user.is_admin,
+            "role": user.role
+        })
```
czue, line 113: if i were nit picking this i'd consider pushing the json generation into a serializer (is that a thing in fastAPI?) or at least a helper function on the user object. it also might be slightly more pythonic to use a list comprehension to generate it. e.g.
```
team_users = [user.to_dict() for user in db_team.users]
```

app/api/teams.py
```diff
@@ -0,0 +1,178 @@
...
+        id=db_team.id,
+        name=db_team.name,
+        email=db_team.email,
+        phone=db_team.phone,
+        billing_address=db_team.billing_address,
+        is_active=db_team.is_active,
+        created_at=db_team.created_at,
+        updated_at=db_team.updated_at,
+        users=team_users
+    )
```
czue, line 126: similar comment to the above in terms of manually constructing the response. I asked claude about this and here's what it said. but, again, might be missing something.

---

No, this manual serialization approach is not considered a best practice in FastAPI, and there are more canonical and maintainable ways to handle this.

FastAPI is built on top of Pydantic, which is specifically designed to handle serialization/deserialization in a clean, type-safe way. The more canonical approach would be to:

1. Define Pydantic models for your response schemas
2. Use model_dump() (or dict() in older versions) for SQLAlchemy model conversion
3. Optionally use SQLAlchemy's declarative models with an ORM config

Here's how I would refactor this code:

```python
from pydantic import BaseModel
from datetime import datetime
from typing import List

# Pydantic models for responses
class UserResponse(BaseModel):
    id: int
    email: str
    is_active: bool
    is_admin: bool
    role: str

    class Config:
        from_attributes = True  # This enables ORM mode in newer Pydantic versions

class TeamWithUsers(BaseModel):
    id: int
    name: str
    email: str
    phone: str
    billing_address: str
    is_active: bool
    created_at: datetime
    updated_at: datetime
    users: List[UserResponse]

    class Config:
        from_attributes = True

@router.get("/{team_id}", response_model=TeamWithUsers)
async def get_team(
    team_id: int,
    current_user: DBUser = Depends(get_current_user_from_auth),
    db: Session = Depends(get_db)
):
    """
    Get a team by ID. Accessible by admin users or users associated with the team.
    """
    # Check if team exists
    db_team = db.query(DBTeam).filter(DBTeam.id == team_id).first()
    if not db_team:
        raise HTTPException(status_code=404, detail="Team not found")

    # Check if user is authorized
    if not current_user.is_admin and current_user.team_id != team_id:
        raise HTTPException(
            status_code=status.HTTP_403_FORBIDDEN,
            detail="Not authorized to access this team"
        )

    # FastAPI will automatically serialize the SQLAlchemy model to the Pydantic model
    return db_team
```

The benefits of this approach are:

1. **Type Safety**: Pydantic models provide runtime type checking
2. **Separation of Concerns**: Your data models (schemas) are separate from your route logic
3. **Automatic Validation**: Pydantic handles validation automatically
4. **Documentation**: FastAPI uses these models to generate OpenAPI documentation
5. **Maintainability**: Changes to the response structure only need to be made in one place
6. **Less Code**: No need for manual field-by-field mapping

You can also use SQLAlchemy models with Pydantic by setting up your SQLAlchemy models to work with Pydantic directly:

```python
from sqlalchemy.orm import declarative_base
from sqlalchemy import Column, Integer, String, Boolean, DateTime

Base = declarative_base()

class DBTeam(Base):
    __tablename__ = "teams"
    
    id = Column(Integer, primary_key=True)
    name = Column(String)
    email = Column(String)
    # ... other fields ...

    # If you want to customize the serialization
    def to_dict(self):
        return {
            "id": self.id,
            "name": self.name,
            # ... other fields ...
            "users": [user.to_dict() for user in self.users]
        }
```

The manual approach you showed might be useful in very specific cases where you need custom serialization logic that can't be handled by Pydantic's configuration options, but for most use cases, leveraging Pydantic models is the recommended way in FastAPI.

app/api/teams.py
```diff
@@ -0,0 +1,178 @@
...
+    """
+    Update a team. Accessible by admin users or team admins.
+    """
+    # Check if team exists
+    db_team = db.query(DBTeam).filter(DBTeam.id == team_id).first()
+    if not db_team:
+        raise HTTPException(status_code=404, detail="Team not found")
+
+    # Check if user is authorized
+    check_team_admin(current_user, team_id)
```
czue, line 146: nit: could avoid a db lookup by moving this check first

czue (commented):
So I kind of just went through and dropped "stream of consciousness" feedback wherever I noticed something. I did not try to educate myself much on FastAPI best-practices so some of my comments might not make sense and just be a reflection of me not knowing how a particular thing is commonly done.

As instructed I did not look at JS / TS code at all. I also didn't look at the tests at all, though let me know if you think I should.

Let me know if this level of review was helpful or if I should invest more / less time into it in the future...

app/api/teams.py
PhiRho: With some playing around I have managed something like a decorator - it is more using FastAPI dependencies, but it serves the same purpose. Definitely cleaner!

app/api/teams.py
PhiRho: This is a "half implementation" since I'm not sending a validation email as part of the registration flow (yet). That needs a little more thought as it brings in new dependencies.

app/db/models.py
PhiRho: Yeah - that was a big part of the design conversation. This speaks to your point about the naming being slightly confusing in that a `Team` is a billable entity. With that in mind, usage has to roll up a single tree and so a user being a member of multiple teams causes a _lot_ of problems.

app/db/models.py
PhiRho: This relationship is confusingly defined because of some issues with the original implementation (it took a while to break some circular references which didn't make sense). It should be a straightforward `back_populates` relationship defined to use the ORM to manage the relationship rather than having to do things multiple times.

app/db/models.py
PhiRho: This ties into the shape of the entity - there may be multiple users with `admin` permissions in a team. In which case we would not have a single point of contact. `admin_email` might make the most sense.

app/api/teams.py
PhiRho: Yup - this should 100% be `Pydantic`

app/api/teams.py
PhiRho: Solved by the decoration approach.

app/api/users.py
```diff
@@ -37,8 +38,39 @@ async def list_users(
     current_user: DBUser = Depends(get_current_user_from_auth),
     db: Session = Depends(get_db)
 ):
-    check_admin(current_user)
-    return db.query(DBUser).all()
+    """
+    List users. Accessible by admin users or team admins for their team members.
+    """
```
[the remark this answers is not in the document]
PhiRho: This is a good thought. I'll save that refactor for a second pass though.

app/api/users.py
```diff
@@ -37,8 +38,39 @@ async def list_users(
...
+    if current_user.is_admin:
+        users = db.query(DBUser).all()
+        # Add team information to each user
+        for user in users:
+            if user.team_id:
+                team = db.query(DBTeam).filter(DBTeam.id == user.team_id).first()
```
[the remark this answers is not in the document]
PhiRho: Whoops - good catch, thank you.

czue (approved):
Only took a quick pass through the recent commits, but the latest changes look great! The `Depends()` thing (and that you can chain them ?!) is quite cool.

PR #35, opened 2025-04-29
Refactor Private AI Keys to allow team owned keys

PhiRho:
This refactor splits creation of LLM tokens and vector DBs as well as allowing AI keys to be owned by teams rather than only by users.

The refactor is necessary because in general usage there are expected to be more LLM tokens than Vector DB credentials. This allows us to not create a large number of entirely unused DBs. We maintain the original database table and API route for backwards compatibility since trial users exist. The refactor also cleans up the `private_ai` class, and ensures the `lite_llm` service is used for all downstream calls. Tests have been added (positive and negative) to ensure no functionality has been removed as new functionality is added.

### DIFF
--- a/app/api/private_ai_keys.py
+++ b/app/api/private_ai_keys.py
@@ -231,12 +231,12 @@ async def list_private_ai_keys(
 async def delete_private_ai_key(
     key_name: str,
     current_user = Depends(get_current_user_from_auth),
+    user_role: str = Depends(get_role_min_key_creator),
     db: Session = Depends(get_db)
 ):
-    # Get the private AI key record
+    # First try to find the key
     private_ai_key = db.query(DBPrivateAIKey).filter(
-        DBPrivateAIKey.database_name == key_name,
-        DBPrivateAIKey.owner_id == current_user.id
+        DBPrivateAIKey.database_name == key_name
     ).first()
 
     if not private_ai_key:
@@ -244,6 +244,35 @@ async def delete_private_ai_key(
             status_code=status.HTTP_404_NOT_FOUND,
             detail="Private AI Key not found"
         )
+
+    # Check if user has permission to delete the key
+    if current_user.is_admin:
+        # System admin can delete any key
+        pass
+    elif user_role == "admin":
+        # Team admin can only delete keys from their team
+        if private_ai_key.team_id is not None:
+            # For team-owned keys, check if it belongs to the admin's team
+            if private_ai_key.team_id != current_user.team_id:
+                raise HTTPException(
+                    status_code=status.HTTP_404_NOT_FOUND,
+                    detail="Private AI Key not found"
+                )
+        else:
+            # For user-owned keys, check if the owner is in the admin's team
+            owner = db.query(DBUser).filter(DBUser.id == private_ai_key.owner_id).first()
+            if not owner or owner.team_id != current_user.team_id:
+                raise HTTPException(
+                    status_code=status.HTTP_404_NOT_FOUND,
+                    detail="Private AI Key not found"
+                )
+    else:
+        # Regular users can only delete their own keys
+        if private_ai_key.owner_id != current_user.id:
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND,
+                detail="Private AI Key not found"
+            )
 
     # Get the region
     region = db.query(DBRegion).filter(DBRegion.id == private_ai_key.region_id).first()
--- a/app/api/private_ai_keys.py
+++ b/app/api/private_ai_keys.py
@@ -5,7 +5,10 @@ from fastapi import status
 import logging
 
 from app.db.database import get_db
-from app.schemas.models import PrivateAIKey, PrivateAIKeyCreate, PrivateAIKeySpend, BudgetPeriodUpdate
+from app.schemas.models import (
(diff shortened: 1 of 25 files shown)


app/core/security.py
```diff
@@ -122,19 +136,27 @@ async def check_system_admin(current_user: DBUser = Depends(get_current_user_fro
...
         )
 
-async def check_team_admin(current_user: DBUser = Depends(get_current_user_from_auth)):
-    """Check if the current user is an admin of a team"""
-    # System admin overrides all
-    if (not current_user.is_admin) and (not (current_user.team_id and current_user.role == "admin")):
+def get_user_role(minimum_role: str, current_user: DBUser):
+    if current_user.is_admin:
+        return "system_admin"
+    elif user_role_hierarchy[current_user.role] > user_role_hierarchy[minimum_role]:
```
czue, line 142: Fine to leave this if it works for now, but long term I wouldn't expect a simple hierarchy of permissions that assumes each level contains all of the levels below it to match real-world use cases. A common way to address that is to set up a roles --> rights map where the rights are checked and the roles grant sets of rights.

app/db/models.py
```diff
@@ -89,7 +89,7 @@ class DBPrivateAIKey(Base):
 
     def to_dict(self):
         return {
-            "id": self.id,
+            "id": self.id or -1,
```
czue, line 92: this magic `-1`  is a bit opaque. could make it a meaningful constant name or add a comment about what it means?

app/db/postgres.py
```diff
@@ -13,11 +13,10 @@ def __init__(self, region: DBRegion = None):
             self.admin_user = region.postgres_admin_user
             self.admin_password = region.postgres_admin_password
             self.port = region.postgres_port
-            self.litellm_api_url = region.litellm_api_url
         else:
             raise ValueError("Region is required for PostgresManager")
 
-    async def create_database(self, owner: str, litellm_token: str, name: str = None, user_id: int = None) -> Dict:
+    async def create_database(self) -> Dict:
```
czue, line 19: super nit picky, but recent python versions can just use `dict` (and `list`) instead of having to work with the `typing` values.

app/api/private_ai_keys.py
```diff
@@ -45,15 +168,87 @@ async def create_private_ai_key(
...
 
     Note: You must be authenticated to use this endpoint.
-    Only admins can create keys for other users.
+    Only admins can create keys for other users or teams.
     """
+    # First create the LiteLLM token
+    llm_token = await create_llm_token(private_ai_key, current_user, user_role, db, store_result=False)
+
+    # Then create the vector database
+    vector_db = VectorDBCreate(
```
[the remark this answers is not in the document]
czue: Sorry for stream of consciousness. I re-read the PR description and this makes more sense now, and I think the main reason this works the way it does is for backwards compatibility?

In which case the argument for re-naming / changing the route is also kind of moot.

czue (commented):
Reviewed pretty fast again but overall looks good! The permissions refactors and pushing stuff into the service are particularly nice.

As before I didn't review the front end code or tests at all, though I noticed there were a lot of tests which is good (assuming they are correct :sweat_smile: )

Just left a couple of drive-by comments. Big picture, the only other feedback I have is that there's still a lot of pretty specific if / else branching logic around permission checking in a few places which is further complicated by the super-admin / team admin / user hierarchy. Not sure if there are any better ways to structure but pushing as much of that as possible into some helper functions (as you've done) and making sure those functions are very well tested seems like a good path to keep going down...

app/api/private_ai_keys.py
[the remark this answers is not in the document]
PhiRho: >  I think the main reason this works the way it does is for backwards compatibility?
> 
> In which case the argument for re-naming / changing the route is also kind of moot.

Basically this. It isn't great, but there may be further opportunities to clean things up a bit as the new routes come into play (I hope).

app/db/postgres.py
PhiRho: 👍🏻

app/api/private_ai_keys.py
```diff
@@ -1,34 +1,157 @@
...
+    """
+    Helper function to determine ownership information based on user role and input.
+    Returns a tuple of (owner_id, team_id).
+    """
+    # If no owner_id or team_id is specified, use the current user's ID
+    if owner_id is None and team_id is None:
+        owner_id = current_user.id
+
+    # Fail fast without having to do DB lookups
+    if user_role in ["key_creator", "user"]:
```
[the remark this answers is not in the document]
PhiRho: I'm playing fast and loose with `Literal` here, but still not quite getting it right :).

app/core/security.py
PhiRho: I am more and more seeing the need to pull out a full permissions "system" which will be more complex but also more flexible and allow for everything up to resource based access control. I want to talk the UX through before I go that route though. We'll call this one a process.

PhiRho:
Updated the workflow to run the tests on dev pull-requests. The frontend tests fail due to not properly existing, so that can be ignored. I'll do a full permissions refactor as its own thing, since I think it will touch a lot of different aspects. Main update here is using the `UserRole` Literal a bit more aggressively to avoid stringly typed parameters.

czue (approved):
Nice, thanks for the updates!

PR #43, opened 2025-05-13
Instrument the amazeeai backend with prometheus metrics

PhiRho:
Adds instrumentation for basic RED metrics, as well as custom tracking on authentication endpoints to keep track of incoming email addresses.

Example dashboard is included (local grafana and prometheus)
image


PhiRho:
Latest version of the default dashboard, taking into account cardinality fixes in the code:
image

app/middleware/auth.py
```diff
@@ -0,0 +1,55 @@
...
+from app.db.database import get_db
+import logging
+from typing import Optional, Dict, Any
+
+logger = logging.getLogger(__name__)
+
+class AuthMiddleware(BaseHTTPMiddleware):
+    async def dispatch(self, request: Request, call_next):
+        # Skip auth for certain paths
+        if request.url.path in ["/health", "/docs", "/openapi.json", "/metrics"]:
```
czue, line 13: super nit picky but could maybe benefit from being extracted to `const.PUBLIC_ENDPOINTS` or similar and then using that here and [here](https://github.com/amazeeio/amazee.ai/pull/43/files#diff-0fdfa731bba127e9566ed41eba69ab3eaa6f3439403d1b457a92d92e7dcf8473R23), just imagining it'd be easier to manage centrally whenever the list changes...

app/middleware/auth.py
```diff
@@ -0,0 +1,55 @@
...
+            if access_token:
+                # Get a fresh database session
+                db = next(get_db())
+                try:
+                    user = await get_current_user_from_auth(
+                        access_token=access_token if access_token else None,
+                        authorization=auth_header if auth_header else None,
+                        db=db
+                    )
+                    # Store essential user data instead of the full SQLAlchemy object
```
czue, line 40: what's the benefit of doing this? especially given that at least sometimes we need the full SQLAlchemy object later (via `get_current_user_from_auth`)

czue (approved):
looks good! keeping in mind that i don't know much about prometheus and didn't review the config files

app/middleware/auth.py
PhiRho: Bug fixing... I got a `DetachedInstanceError` in some of my tests due to lazy loading some of the fields on the user later on. We do still have to reload the SA object later, but it is fewer times than it was.

PR #56, opened 2025-06-02
Begin the Stripe Integration

PhiRho:
This is a fairly large change to begin the integration with Stripe for billing purposes. Once this is in we still need to refine things, but that can be done in smaller chunks.

Key changes in this PR:
- add billing routes needed for creating sessions and customers
- add product routes to allow administrators to manage products
- add webhook management for Stripe events
- add the concept of resource limits based on associated products
- add basic FE for these pieces


czue (commented):
As always i left mostly drive-by commentary and didn't review front end or tests closely.

I think my biggest pieces of feedback are:

1. Maybe try to reduce DB queries (especially in the resource limit checks)
2. Maybe make the Stripe webhook use cases you expect to deal with the various business scenarios (new sub, renewal, upgrade, downgrade, cancellation) a bit more explicit.

In terms of your questions from Slack:

> Should the background processor for webhook events stay in the API file, or move to the worker file?

I'd vote move them.

> I'd prefer to have the Webhook endpoint configured dynamically so that the signing secret is never exposed (outside of a local testing environment), does that fit with normal practice, or should I just accept the manual step?

I think either is fine. I've seen apps put it in an env var or in the DB. The Stripe private API key is a much more important secret than the webhook signing secret.

app/api/billing.py
```diff
@@ -0,0 +1,271 @@
...
+            # We assume that the invoice is related to a subscription
+            subscription = event_object.parent.subscription_details.subscription
+            product_id = await get_product_id_from_subscription(subscription)
+            await apply_product_for_team(db, customer_id, product_id)
+        elif event_type in subscription_success_events:
+            product_id = await get_product_id_from_subscription(event_object.id)
+            await apply_product_for_team(db, customer_id, product_id)
+        elif event_type in session_success_events:
+            product_id = await get_product_id_from_session(event_object.id)
+            await apply_product_for_team(db, customer_id, product_id)
```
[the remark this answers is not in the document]
PhiRho: > a single event i listen to for new subs and then a single one for renewals

I would prefer that! I may have read too much of the documentation... I think I have managed to whittle it down. Now it comes down to testing.

app/services/stripe.py
```diff
@@ -0,0 +1,310 @@
...
+            # First delete the old endpoint
+            stripe.WebhookEndpoint.delete(existing_endpoint.id)
+            logger.info(f"Deleted existing webhook endpoint: {existing_endpoint.id}")
+
+        # Create new webhook endpoint
+        endpoint = stripe.WebhookEndpoint.create(
+            url=webhook_url,
+            enabled_events=[
+                "checkout.session.completed",
+                "customer.subscription.deleted"
```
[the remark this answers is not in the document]
PhiRho: Me forgetting what I was doing because the change was too big. Good catch. I need to go through and remove things which aren't actually going to be used in the first iteration, as well as fix up some other aspects.

app/api/private_ai_keys.py
```diff
@@ -390,6 +412,67 @@ async def list_private_ai_keys(
...
+
+    try:
+        # Get LiteLLM key info
+        litellm_data = await litellm_service.get_key_info(private_ai_key.litellm_token)
+        info = litellm_data.get("info", {})
+        logger.info(f"LiteLLM key info: {info}")
+
+        # Combine database key info with LiteLLM info
+        key_data = private_ai_key.to_dict()
+        key_data.update(info)
```
czue, line 466: I was looking at [this commit](https://github.com/amazeeio/amazee.ai/pull/56/commits/3ba1d8e09164163f412209cd49670f3239658bde#diff-5de65c6fdfae351c3d108d7689ccd69c3701f6b8ce3a4f9e2ff5f9d7ce3e3258R466) and noticed that it previously changed "expires" to "expires_at". just flagging in case that's a bug/problem.

app/core/resource_limits.py
```diff
@@ -0,0 +1,199 @@
...
+        DBProduct.id == DBTeamProduct.product_id
+    ).outerjoin(
+        DBPrivateAIKey,
+        or_(
+            DBPrivateAIKey.team_id == DBTeam.id,
+            DBPrivateAIKey.owner_id.in_(
+                db.query(DBUser.id).filter(DBUser.team_id == DBTeam.id)
+            )
+        )
+    ).first()
```
czue, line 95: you really ran with the sql optimization! :joy: :dizzy_face:

app/services/stripe.py
```diff
@@ -0,0 +1,257 @@
...
+# Full list of possible events: https://docs.stripe.com/api/events/types
+invoice_success_events = ["invoice.paid"] # Renewal
+subscription_success_events = ["customer.subscription.resumed", "customer.subscription.created"] # New subscription
+session_failure_events = ["checkout.session.async_payment_failed", "checkout.session.expired"] # Checkout failure
+subscription_failure_events = ["customer.subscription.deleted", "customer.subscription.paused"] # Subscription failure
+invoice_failure_events = ["invoice.payment_failed"] # Invoice failure
+
+success_events = invoice_success_events + subscription_success_events
+failure_events = session_failure_events + subscription_failure_events + invoice_failure_events
+known_events = success_events + failure_events
```
czue, line 24: nit: it might be slightly more pythonic to name these in `CONSTANT_CASE`

czue (approved):
Reviewed quite a bit quicker this time, but looks good! Those new DB queries are quite something...

PR #73, opened 2025-06-13
Re-apply JWT and bugfix periodic job

PhiRho:
All the changes for the JWT, the periodic job, extending trials, etc.


PhiRho:
I messed up the git-history by accidentally going to main before dev. As of this merge the two should be back to parity allowing ordinary development to proceed.

PR #76, opened 2025-06-13
Free stuff

PhiRho:
Make it possible for a system admin to configure a specific pricing table to offer free-stuff to people.


app/main.py
```diff
@@ -15,11 +15,15 @@
...
 from apscheduler.triggers.cron import CronTrigger
-
 import os
 import logging
+import fcntl
+import tempfile
 from datetime import UTC
 
+# Set timezone environment variable to prevent tzlocal warning
+os.environ['TZ'] = 'UTC'
```
czue, line 25: nit: guessing it's fine but this feels a bit weird (I'm not familiar with this error so don't have an alternate proposal).

app/main.py
```diff
@@ -40,11 +44,30 @@ async def lifespan(app: FastAPI):
     scheduler = AsyncIOScheduler()
 
     async def monitor_teams_job():
+        # Create a lock file in the temp directory
+        lock_file = os.path.join(tempfile.gettempdir(), 'monitor_teams.lock')
```
czue, line 48: what's the purpose of the lock here? do we ever have to worry about this job running on multiple machines (if so, a file-based approach might not work). separately, i wonder if there's a library that might support this more reusably/reliably

czue (approved):
had a couple small things but looks good. didn't review tests or front end code

app/main.py
PhiRho: The scheduled task sends an email in certain cases. At the moment, because of the way `uvicorn` is started it sends four of them. I will look at if there is a library which can do the locking with a DB table rather, but will do that when I add certain other resilience aspects (like what happens if the run is interrupted). This is sufficient for now.

PR #78, opened 2025-06-24
Pricing UX improvements

PhiRho:
- Send a new URL with token if previous was expired
- Ensure we only validate once to avoid any duplicate emails


czue (commented):
Looks good. Only question was whether you checked if there's a third party library that can handle DB locking for you? It seems like the type of thing that might exist, and also the type of thing with enough edge cases that it might be useful to offload it to a supported package.

PR #83, opened 2025-06-27
update(resource_limits): increase default resource limits for users and keys

PhiRho:
- Adjust default user count from 1 to 100.
- Increase keys per user from 1 to 5.
- Raise total keys from 2 to 500.
- Update service keys from 1 to 100.
- Expand vector DB count from 1 to 100.


bonus414:
@PhiRho 
The updated limits look good to me.
I see there are failing checks, and since the errors appear to be unrelated to this PR’s scope and I’m not the right person to debug them, I don't know that I should just approve.

PR #99, opened 2025-07-22
Dedicated

PhiRho:
- Allow regions to be marked as dedicated
- Add validation of the DB and LLM endpoints on creation
- Add association between regions and teams
- Minor refactoring of region management
- Overall test coverage improvement for region management


app/api/regions.py
```diff
@@ -1,29 +1,100 @@
...
+        bool: True if validation succeeds, raises HTTPException if it fails
+    """
+    try:
+        # Attempt to connect to the database
+        conn = await asyncpg.connect(
+            host=host,
+            port=port,
+            user=user,
+            password=password
+        )
```
czue, line 75: assuming this doesn't need to check any permissions? e.g. by making a select statement or similar.

edit: I see it does attempt a query shortly after this is called, so that's probably fine.

app/api/regions.py
```diff
@@ -51,24 +133,40 @@ async def list_regions(
...
+        DBRegion.is_active == True,
+        DBRegion.is_dedicated == True,
+        DBTeamRegion.team_id == current_user.team_id
+    ).all()
 
-@router.get("/admin", response_model=List[Region])
+    non_dedicated_regions = db.query(DBRegion).filter(
+        DBRegion.is_active == True,
+        DBRegion.is_dedicated == False
+    ).all()
```
czue, line 157: nit: you could use the same variable for the "regular user" return value above to avoid duplicating the business logic.

app/api/regions.py
```diff
@@ -164,4 +249,126 @@ async def update_region(
...
+
+    # Remove the association
+    db.delete(association)
+
+    try:
+        db.commit()
+    except Exception as e:
+        db.rollback()
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
```
czue, line 342: nit: should this be a 500 instead of a 404?

app/api/regions.py
```diff
@@ -164,4 +249,126 @@ async def update_region(
...
+        region_id=region_id
+    )
+    db.add(team_region)
+
+    try:
+        db.commit()
+    except Exception as e:
+        db.rollback()
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
```
czue, line 308: nit: should this be a 500 instead of a 404?

czue (approved):
Looks good, just had some tiny stuff.

The "dedicated" terminology threw me off a bit, since it seems like multiple teams can share a "dedicated" region. I don't think that's a big deal if it makes sense to you and the people using it though.

PR #117, opened 2025-08-22
Merge teams

PhiRho:
- Add functionality for merging teams
  - Source team must not have a product
  - Duplicate keys are renamed or deleted
  - Users are moved
  - LiteLLM is updated
- Add UI functionality
- Potentially fix "expired" badges
- Keep system admins logged in longer

MergeTeams

### DIFF
--- a/app/api/teams.py
+++ b/app/api/teams.py
@@ -1,20 +1,22 @@
 from fastapi import APIRouter, Depends, HTTPException, status
 from sqlalchemy.orm import Session
-from typing import List
+from typing import List, Optional
 from datetime import datetime, UTC
 import logging
 
 from app.db.database import get_db
-from app.db.models import DBTeam, DBTeamProduct, DBUser
+from app.db.models import DBTeam, DBTeamProduct, DBUser, DBPrivateAIKey, DBRegion
 from app.core.security import check_system_admin, check_specific_team_admin, get_current_user_from_auth
 from app.schemas.models import (
     Team, TeamCreate, TeamUpdate,
-    TeamWithUsers
+    TeamWithUsers, TeamMergeRequest, TeamMergeResponse
 )
 from app.core.resource_limits import DEFAULT_KEY_DURATION, DEFAULT_MAX_SPEND, DEFAULT_RPM_PER_KEY
 from app.services.litellm import LiteLLMService
 from app.services.ses import SESService
 from app.core.worker import get_team_keys_by_region, generate_pricing_url, get_team_admin_email
+from app.api.private_ai_keys import delete_private_ai_key
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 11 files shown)


app/api/teams.py
```diff
@@ -219,3 +221,219 @@ async def extend_team_trial(
...
+                        db=db
+                    )
+                except Exception as e:
+                    logger.error(f"Failed to delete key {key.id}: {str(e)}")
+                    # Continue with other keys even if one fails
+
+        return remaining_keys
+    elif strategy == "rename":
+        # Rename conflicting keys in team2
+        suffix = rename_suffix or "_merged"
```
czue, line 262: probably not a real issue, but do we have to worry about the renamed key also having a conflict? like if the other team already had a key named "key1_merged"?

app/api/teams.py
```diff
@@ -219,3 +221,219 @@ async def extend_team_trial(
...
+        # Delete conflicting keys from database if db session provided
+        if db and current_user:
+            for key in keys_to_delete:
+                try:
+                    await delete_private_ai_key(
+                        key_id=key.id,
+                        current_user=current_user,
+                        user_role="system_admin",  # System admin context for merge operations
+                        db=db
+                    )
```
czue, line 254: is there any scenario where these get deleted but then for whatever reason the save/commit call fails when we actually do the migration and then the keys are gone?

app/api/teams.py
```diff
@@ -219,3 +221,219 @@ async def extend_team_trial(
...
+        # Update LiteLLM key associations for each region
+        for region_id, region_keys in keys_by_region.items():
+            # Get region info
+            region = db.query(DBRegion).filter(
+                DBRegion.id == region_id,
+                DBRegion.is_active == True
+            ).first()
+
+            if not region:
+                continue
```
czue, line 399: when would this be the case? and is it a problem if there are relevant keys associated with an inactive region?

app/api/teams.py
```diff
@@ -219,3 +221,219 @@ async def extend_team_trial(
...
+                api_url=region.litellm_api_url,
+                api_key=region.litellm_api_key
+            )
+
+            # Update team association for each key in this region
+            for key in region_keys:
+                try:
+                    await litellm_service.update_key_team_association(
+                        key.litellm_token,
+                        f"{region.name.replace(' ', '_')}_{target_team.id}"
```
czue, line 412: this is just like a magic string that is parsed to back out the associated team id?

app/api/teams.py
```diff
@@ -219,3 +221,219 @@ async def extend_team_trial(
...
+    target_team_id: int,
+    db: Session
+) -> int:
+    """Migrate users from source team to target team"""
+    migrated_count = 0
+    for user in users:
+        if user.team_id != target_team_id:
+            user.team_id = target_team_id
+            migrated_count += 1
+    db.commit()
```
czue, line 284: it might be fine but it seems like these interspersed commits could result in the DB getting into an inconsistent state if something fails partway through.

app/api/teams.py
```diff
@@ -219,3 +221,219 @@ async def extend_team_trial(
...
+            conflicts_resolved=conflicts if conflicts else None,
+            keys_migrated=keys_migrated,
+            users_migrated=users_migrated
+        )
+
+    except HTTPException:
+        # Re-raise HTTP exceptions as-is
+        raise
+    except Exception as e:
+        db.rollback()
```
czue, line 434: will this actually work if commits have been made in the `_migrate_keys_to_team` and `_migrate_users_to_team` functions?

czue (commented):
Looks good. Only possible question is arround commit/rollback behavior in various places.

(I didn't review tests or front end  code)

app/api/teams.py
PhiRho: Good catch - let me change the default to include the team it is coming from (which would be unique).

app/api/teams.py
PhiRho: There _is_, but I don't have a good way to rollback in that case since this delete is fully destructive including tearing down vector DBs if they are part of the key. Given that the keys being deleted is the expected behaviour I'm OK with the risk.

PR #125, opened 2025-09-04
Sales dashboard

PhiRho:
Implements a bunch of changes necessary for showing a sales dashboard to specific types of systems users

## UI
- Add the sales-dashboard in two places. Both system admin and sales users can see it
- Ensure this dashboard has sorting and filters
- Ensure appropriate columns are displayed

image
image
image

## Backend
- Add a new API specifically to pull the exact data needed for this dashboard. 
- Refactor role based access control to allow adding `sales` as a role
- Use RBAC to restrict the sales dashboard API to sales and system admin users
- Test coverage

### DIFF
--- a/app/api/billing.py
+++ b/app/api/billing.py
@@ -4,7 +4,7 @@ import logging
 import os
 from datetime import datetime, UTC
 from app.db.database import get_db
-from app.core.security import check_specific_team_admin, check_system_admin
+from app.core.security import get_role_min_specific_team_admin, get_role_min_system_admin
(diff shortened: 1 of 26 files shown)


czue (approved):
The other way I've seen this type of things done is by splitting permissions into roles and rights. So there is more of a logical hierarchy that says "key management requires the key management right" and then "the key management right belongs to the system admin, team admin, and key manager roles". Then when you do the permission check you instead just check if the role has the right.

Probably not worth a refactor at this stage, but it does make things more declarative and avoid having to have lots of helper functions that combine lists of roles in ways that might be more prone to errors....

PR #141, opened 2025-10-02
Feature/frontend updates

bonus414:
This PR updates the create-ai-key-dialog.tsx component with frontend adjustments.

On this page: https://dashboard.amazee.ai/admin/private-ai-keys when you click "Create Key" and the modal appears you can go to "Assign to" and then you can't actually search and it's a bit annoying. So I worked with Claude on a fix. That being said, Docker is just not wanting to run on my computer and I've reached my frustration point with it. So I haven't been able to test this at all.

Can you test this and see if this now allows us to search and probably need to look at the code because this is definitely AI generated?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ ENV/
 *.swp
 *.swo
 .cursor/
+.cursorrules
 
 # OS generated files
 .DS_Store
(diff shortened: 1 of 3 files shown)


PhiRho:
image

Once I got to the bottom of "why are the tests failing" I could also get through the login error to confirm the change in lagoon. LGTM.

PR #154, opened 2025-10-25
feat: add database name filter to private AI keys admin page

bonus414:
- Add database name filter field next to existing filters
- Implement client-side filtering logic for database names
- Update clear filters functionality to include database name filter
- Update hasActiveFilters check to include database name filter
- Filter matches database names case-insensitively as user types


bonus414:
@PhiRho it's low risk enough we can just go straight to prod 😄

PR #158, opened 2025-11-06
chore: add .python-version to locally override python version to 3.12

dan2k3k4:
- keep python version the same version as dockerfile


dan2k3k4:
FYI this can also be achieved with `$PYENV_VERSION` so we could add a `.env.example` file and `.envrc`, which would be executed by the `direnv` tool, and can run a `cp .env.example .env` if the file doesn't exist locally, that just outputs `PYENV_VERSION=3.12`

So if we prefer to do it that way, then I can adjust this PR. As `.python-version` might be read by the deployed Lagoon env, whereas the `.env.example` workflow won't copy to a `.env` file as `direnv` usually isn't installed on the Lagoon environments.

PR #159, opened 2025-11-06
Feature/invitation only passwordless auth

bonus414:
I attempted to update some guardrails for access to the frontend of this. if we're going to continue to just tell people they will have this "dashboard" then I need some control. This seemed like a good MVP for now. I definitely need it tested, reviewed, and all the things!

### DIFF
--- a/docs/plans/2025-11-06-invitation-only-passwordless-auth-design.md
+++ b/docs/plans/2025-11-06-invitation-only-passwordless-auth-design.md
@@ -1,0 +1,159 @@
+# Invitation-Only Passwordless Authentication Design
+
+**Date:** 2025-11-06
+**Status:** Approved
+**Branch:** `feature/invitation-only-passwordless-auth`
+
+## Problem Statement
+
+The current authentication system allows anyone to self-register by requesting a verification code. The `/sign-in` endpoint auto-creates users and teams if they don't exist, creating a cost and abuse risk - anyone who discovers the dashboard URL can create free LLM keys until detected.
+
+## Goals
+
+1. Remove password-based authentication (verification code only)
+2. Implement invitation-only access (admin must pre-create accounts)
+3. Prevent self-service registration while maintaining simple customer experience
+4. Keep admin workflow simple with minimal commands
+
+## Solution: Option A - Whitelist Approach
+
+### Architecture Changes
+
+#### Frontend Changes
+
+**File: `/auth/login/page.tsx`**
+- Show only passwordless login form
+- Remove password login option entirely
+
+**File: `components/auth/passwordless-login-form.tsx`**
+- Remove "Sign in with password" button and divider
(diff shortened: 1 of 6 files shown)


PhiRho:
OK, so we want users to only access the dashboard if they've been invited, that makes sense. There are a couple of important points to watch out for.
1. The current passwordless auth flow which creates a team etc is what we use in the Drupal Module - that's the actual reason it is that shape.
2. An alternative solution would be to expand the RBAC and permissions design to allow a user access to the dashboard, then the invite would add that permission, and we can block the dashboard access for a user before even completing the sign-in flow.

docs/plans/2025-11-06-invitation-only-passwordless-auth-design.md
```diff
@@ -0,0 +1,159 @@
...
+The current authentication system allows anyone to self-register by requesting a verification code. The `/sign-in` endpoint auto-creates users and teams if they don't exist, creating a cost and abuse risk - anyone who discovers the dashboard URL can create free LLM keys until detected.
+
+## Goals
+
+1. Remove password-based authentication (verification code only)
+2. Implement invitation-only access (admin must pre-create accounts)
+3. Prevent self-service registration while maintaining simple customer experience
+4. Keep admin workflow simple with minimal commands
+
+## Solution: Option A - Whitelist Approach
```
PhiRho: Note to self, LLMs need to learn to use allowlist and blocklist.

docs/plans/2025-11-06-invitation-only-passwordless-auth-design.md
```diff
@@ -0,0 +1,159 @@
...
+curl -X POST https://api.amazee.ai/users \
+  -H "Authorization: Bearer YOUR_ADMIN_TOKEN" \
+  -H "Content-Type: application/json" \
+  -d '{"email": "customer@company.com", "password": null, "role": "user"}'
+
+# Step 2: Send verification code to customer
+curl -X POST https://api.amazee.ai/auth/validate-email \
+  -H "Content-Type: application/json" \
+  -d '{"email": "customer@company.com"}'
+
```
PhiRho: Since we're modifying the code anyway, we can use a variation on what we've done for the upgrade path JWT - essentially a short lived magic link, no code required.

PhiRho (changes requested):
I like the idea. I think we need to make sure we can do it without impacting drupal or polydock. This might mean only allowing sign in via magic link to the dashboard, and keeping people logged in longer or something.

bonus414:
@PhiRho 
What's the next step then? (my "I'm not an engineer" side is showing.

bonus414:
Noting for posterity
Nov 10, 2025
Decided against implementing security changes now
- Too complex given limited frontend engineering resources
- Multiple entry points make consistent friction difficult
- Will review risks with Head of Engineering and AI Staff Engineer to decide next steps

PR #162, opened 2025-11-13
chore: add network defaults for services in docker-compose

dan2k3k4:


tobybellwood:
Not sure whether this will work locally with Pygmy as the amazeeio-network is hardcoded in the haproxy config, would be interested to see though?

https://github.com/pygmystack/pygmy/blob/main/internal/service/docker/haproxy/haproxy.go

dan2k3k4:
No, I did not have `pygmy` running, but I wonder if it makes sense to apply these changes if we don't need `pygmy` to run it

Or should we apply the changes and make `pygmy` a requirement?

We only depend on one `uselagoon` image inside the `Dockerfile`
`FROM uselagoon/python-3.12:latest`

PR #172, opened 2025-12-03
chore: fix linting with `ruff check`

dan2k3k4:


tests/test_ses.py
```diff
@@ -63,11 +91,41 @@ def test_create_or_update_template_new(mock_sts_client):
...
+
+        # Mock the templates directory and file
+        mock_templates_dir = MagicMock()
+        service.templates_dir = mock_templates_dir
+
+        mock_file = MagicMock()
+        mock_file.exists.return_value = True
+        mock_file.read_text.return_value = markdown_content
+        mock_templates_dir.__truediv__.return_value = mock_file
+
```
PhiRho: Why is all of this necessary?

PR #214, opened 2025-12-23
chore: add trial recon job to deactivate over-budget trial users

dan2k3k4:
Review feedback on the trial recon job PR addressing three items: missing `db.refresh()` in test fixtures, per-user N+1 query, and inline deactivation logic.

## Changes

- **Test fixtures** — added `db.refresh()` after every `db.commit()` in all five fixtures (`trial_team`, `trial_user`, `trial_region`, `trial_key`, `user_budget_limit`), consistent with `conftest.py` pattern
- **Batch query** — replaced per-user `DBLimitedResource` lookup inside the loop with a single query + dict lookup:
  ```python
  user_limits = db.query(DBLimitedResource).filter(
      DBLimitedResource.owner_type == OwnerType.USER,
      DBLimitedResource.owner_id.in_([user.id for user in users]),
      DBLimitedResource.resource == ResourceType.BUDGET
  ).all()
  user_limit_map = {limit.owner_id: limit for limit in user_limits}
  ```
- **`deactivate_trial_user()` helper** — extracted disable-user + expire-keys logic into a reusable `async def deactivate_trial_user(db, user)`, keeping `monitor_trial_users` focused on orchestration

This PR adds a trial reconciliation job that deactivates trial users after they exhaust their budget. The main changes are:

- Monthly Lagoon cron entries for the trial monitor job.
- A trigger script with database locking around the job run.
- Worker helpers to mark over-budget trial users inactive and expire their LiteLLM keys.
- Tests for in-budget users, over-budget users, and admin exclusion.
</details>

This looks safe to merge.

- No blocking issues found in the changed code.

</details>

| Filename | Overview |
|----------|----------|
| .lagoon.yml | Adds the scheduled trial monitor job for the configured environments. |
| scripts/trigger_trial_recon_job.py | Adds the trial recon trigger with the lock release handled in the inner cleanup path. |
| app/core/worker.py | Adds trial-user monitoring, batched budget-limit lookup, deactivation, and key expiry helpers. |
| tests/test_monitor_trial_users.py | Adds coverage for trial-user monitoring behavior and refreshes committed fixtures. |

</details>

<sub>Reviews (4): Last reviewed commit: ["chore: add trial recon job to deactivate..."](https://github.com/amazeeio/amazee.ai/commit/199c62c85faf90613a6ade60b4aa71f3895d64e0) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=35168860)</sub>


app/core/worker.py
```diff
@@ -1137,3 +1137,62 @@ def generate_pricing_url(admin_email: str, validity_hours: int = 24) -> str:
...
+        trial_team = db.query(DBTeam).filter(DBTeam.admin_email == settings.AI_TRIAL_TEAM_EMAIL).first()
+        if not trial_team:
+            logger.info("Trial team not found, skipping")
+            return
+
+        # Get all active users in the trial team (excluding admin)
+        users = db.query(DBUser).filter(
+            DBUser.team_id == trial_team.id,
+            DBUser.is_active,
+            DBUser.role == "user"
```
alagoa: **thought (separate PR):** could be worth to have an enum for `DBUser.role`.

app/core/worker.py
```diff
@@ -1137,3 +1137,62 @@ def generate_pricing_url(admin_email: str, validity_hours: int = 24) -> str:
...
+                        if key.litellm_token and key.region:
+                            try:
+                                litellm_service = LiteLLMService(
+                                    api_url=key.region.litellm_api_url,
+                                    api_key=key.region.litellm_api_key
+                                )
+                                await litellm_service.update_key_duration(key.litellm_token, "0d")
+                                logger.info(f"Set duration to 0d for key {key.id}")
+                            except Exception as e:
+                                logger.error(f"Failed to expire key {key.id}: {e}")
```
alagoa: **nitpick:** this could be put into a `soft_delete_user()`/`deactivate_user()` method for reusability.

alagoa (approved):
Sorry for the late review, need to figure out my GH notifications 🤓 

Mostly just questions to make sure I understand why we're doing what we are doing here, no major blockers on my side.

Some more doubts (related and unrelated) that came up during review:
1. This job is not being scheduled in `main`, is that coming on a later PR?
2. From what I've seen this does not clash with the other recon jobs we have running, right?
2a. The main monitor only updates `PRODUCT` or `SYSTEM` limits and the trial limits are marked as `MANUAL`.
2b. Is the trial team exempt from the soft delete that runs on the main monitor? Even if it isn't explicitly, it probably is in practice, since the team soft delete requires all users of a team to be quiet for >76 days.

tests/test_monitor_trial_users.py
```diff
@@ -0,0 +1,145 @@
...
+@pytest.fixture
+def trial_team(db: Session):
+    team = DBTeam(
+        name="AI Trial Team",
+        admin_email=settings.AI_TRIAL_TEAM_EMAIL,
+        is_active=True
+    )
+    db.add(team)
+    db.commit()
+    return team
```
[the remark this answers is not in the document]
dan2k3k4: @copilot apply changes based on [this feedback](https://github.com/amazeeio/amazee.ai/pull/214#discussion_r3046610044)

dspachos (commented):
### 🔴 High: Exception swallowed in monitor_trial_users

 ```python
   # worker.py line ~1442
   except Exception as e:
       logger.error(f"Error in trial user monitoring: {e}")
       db.rollback()
       # ⬆️ exception is NOT re-raised
 ```

 Other worker jobs in this file re-raise (raise e). The trigger script expects monitor_trial_users to raise on failure (it has its own try/except/raise around
 the call), but the exception will never propagate — the script will always log "completed successfully" even when the job failed. Should add raise after
 db.rollback().

 ### 🟡 Medium: External side effects before commit (ordering issue)

 In deactivate_trial_user(), LiteLLM keys are expired via the external API before the DB commit happens in monitor_trial_users. If the commit fails:
 - Keys are already expired in LiteLLM
 - But the user remains active in the DB
 - Inconsistent state

 Suggestion: Either commit the user deactivation first, then expire keys (accepting that key expiry might fail but can be retried), or at minimum add a comment
 acknowledging the trade-off.

 ### 🟡 Medium: Missing DBTeam.is_active filter

 ```python
   trial_team = db.query(DBTeam).filter(
       DBTeam.admin_email == settings.AI_TRIAL_TEAM_EMAIL
   ).first()
 ```

 Copilot flagged this — elsewhere (e.g. app/api/auth.py:745) the trial team lookup also filters on is_active. If the trial team is ever deactivated, this job
 would still process its users. Add DBTeam.is_active == True to the filter.

dspachos (approved):
Hi @dan2k3k4 , approved but it seems that `trigger_trial_recon_job.py` is not registered as a cron job (unlike other worker scripts), so trial reconciliation won’t run automatically in dev/prod. Is this correct?

scripts/trigger_trial_recon_job.py
```diff
@@ -0,0 +1,85 @@
...
+    except Exception as e:
+        logger.error(f"Error in trial recon job trigger: {str(e)}")
+        # Try to release lock in case of error
+        try:
+            release_lock(lock_name, db)
+            logger.info("Released lock after error")
+        except Exception as release_error:
+            logger.error(f"Error releasing lock: {str(release_error)}")
+        raise
+    finally:
```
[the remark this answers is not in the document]
dan2k3k4: Fixed in 199c62c85faf — removed the duplicate release_lock in the outer except; the inner finally is the single release point (matches trigger_hard_delete_job.py).

app/core/worker.py
dan2k3k4: Agreed — deferred to a separate PR (DBUser.role enum), resolving here.

PR #254, opened 2026-01-28
feat: forgot password

dan2k3k4:


dspachos (commented):
Hi @dan2k3k4 

Can you please take a look on the following?

1. Missing code expiration in DynamoDB (app/api/auth.py:512)
   - Code stored without TTL, relies on token expiration
   - Consider adding TTL to DynamoDB write

2. Test regression (tests/test_auth.py:848)
   - Changed from 1800 (30 min) to 28800 (8 hours)
   - Needs explanation 

3. Email template inconsistency
   - Template says "valid for 10 minutes" (app/templates/reset-password.md:10)
   - But code has no explicit TTL, only token expires in 1 hour

4. Missing rate limiting
   - /auth/forgot-password endpoint vulnerable to abuse
   - Consider adding rate limiting

PR #256, opened 2026-01-30
chore: add region label

dan2k3k4:


app/db/models.py
```diff
@@ -52,6 +52,7 @@ class DBRegion(Base):
 
     id = Column(Integer, primary_key=True, index=True)
     name = Column(String, unique=True, index=True)
+    label = Column(String, nullable=True)
```
dspachos: - Frontend has required attribute on label input at frontend/src/app/admin/regions/page.tsx:385, 576

- Backend allows nullable (nullable=True)

Users can't create regions without label in UI, but API accepts it (unless I'm missing something)

PR #266, opened 2026-02-04
chore: redirect user if they get logged out, expired login state

dan2k3k4:


frontend/src/utils/api.ts
```diff
@@ -26,6 +26,9 @@ export async function fetchApi(endpoint: string, options: RequestInit = {}) {
   const response = await fetch(url, fetchOptions);
 
   if (!response.ok) {
+    if (response.status === 401 && typeof window !== 'undefined' && !window.location.pathname.startsWith('/auth/login')) {
+      window.location.href = '/auth/login?expired=true';
+    }
     const error = await response.json().catch(() => ({ message: response.statusText }));
```
dspachos: - The code redirects with window.location.href but immediately throws an error after
- This means the redirect starts, but the calling code still receives an error that might be caught and displayed, potentially showing both a redirect and an error toast
- Should either not throw the error after redirecting, or ensure the redirect completes before throwing

PR #269, opened 2026-02-06
chore: add security policy file

dan2k3k4:


smlx (approved):
Looks good! The only thing that you might also want to mention is that private vulnerability reporting is enabled for this repository so users can also use the form at https://github.com/amazeeio/amazee.ai/security

https://docs.github.com/en/code-security/how-tos/report-and-fix-vulnerabilities/report-a-vulnerability/privately-reporting-a-security-vulnerability

PR #281, opened 2026-02-17
chore: require auth for register team related endpoints

dan2k3k4:


smlx (approved):
I'm not very familiar with the design of the API but if admins are the only ones who are supposed to be able to register teams then LGTM.

PR #296, opened 2026-02-25
chore: enable key search by db username or db name

dan2k3k4:


frontend/src/app/admin/private-ai-keys/_components/user-filter.tsx
```diff
@@ -56,19 +56,24 @@ export function UserFilter({ selectedUser, onUserSelect }: UserFilterProps) {
     }
   }, [searchError, toast]);
 
+  const handleClear = () => {
+    onUserSelect(null);
+    setSearchTerm("");
```
dspachos: Perhaps set also: `setIsUserSearchOpen(false);` - if we want the search popover to close (?)

PR #374, opened 2026-04-01
chore: move cronjobs / scheduler to lagoon cron

dan2k3k4:
- [x] Remove unused `try_acquire_lock`/`release_lock` imports from `app/main.py`
- [x] Remove unused `get_db` import from `app/main.py` (also left unused after scheduler removal)
- [x] Fix double-release race in `scripts/trigger_sync_pool_budgets_job.py` — removed the outer `except` block's redundant `release_lock` call so the lock is only released in one place (the inner `finally`)
- [x] Exit with status 0 (not 1) in `scripts/trigger_sync_pool_budgets_job.py` when skipping due to lock held by another process

### DIFF
--- a/scripts/trigger_recon_job.py
+++ b/scripts/trigger_recon_job.py
@@ -55,12 +55,6 @@ async def trigger_recon_job():
 
     except Exception as e:
         logger.error(f"Error in recon job trigger: {str(e)}")
-        # Try to release lock in case of error
-        try:
-            release_lock(lock_name, db)
-            logger.info("Released lock after error")
-        except Exception as release_error:
-            logger.error(f"Error releasing lock: {str(release_error)}")
         raise
     finally:
         db.close()
@@ -81,7 +75,8 @@ def main():
             logger.info(
                 "⚠️  Recon job could not be executed (lock held by another process)"
             )
-            sys.exit(1)
+            # Exiting with status 0 because this is a normal, expected condition (lock held)
+            sys.exit(0)
 
     except Exception as e:
         logger.error(f"❌ Script failed: {str(e)}")
--- a/.lagoon.yml
+++ b/.lagoon.yml
@@ -2,6 +2,19 @@ docker-compose-yaml: docker-compose.yml
 
 environments:
   prod:
+    cronjobs:
+      - name: monitor-teams
+        schedule: "0 * * * *"
+        command: python scripts/trigger_recon_job.py
+        service: backend
+      - name: hard-delete-teams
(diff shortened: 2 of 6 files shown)


dan2k3k4:
@copilot apply changes based on the comments in [this thread](https://github.com/amazeeio/amazee.ai/pull/374#pullrequestreview-4043992657)

alagoa (commented):
Looks good! Just a couple of questions before we ship it

- Doesn't the lock issue copilot flagged apply to the other cronjobs we are lagoonizing on this PR? both `trigger_recon_job` and `trigger_hard_delete_job` followed the same pattern
- Do we need this to run in the dev environments as well? If so, we should either make the project highly available or making it no idle, if that's not done already.

dan2k3k4:
This PR is to make the prod instance of the project HA (highly available)
Non-prod envs (dev, main [stage]) instances are already "no idle"

dan2k3k4:
> Looks good! Just a couple of questions before we ship it
> 
> * Doesn't the lock issue copilot flagged apply to the other cronjobs we are lagoonizing on this PR? both `trigger_recon_job` and `trigger_hard_delete_job` followed the same pattern
> * Do we need this to run in the dev environments as well? If so, we should either make the project highly available or making it no idle, if that's not done already.

@copilot verify if we need to change anything with the first point for the lock issue you flagged

scripts/trigger_recon_job.py
```diff
@@ -81,7 +75,8 @@ def main():
             logger.info(
                 "⚠️  Recon job could not be executed (lock held by another process)"
             )
-            sys.exit(1)
+            # Exiting with status 0 because this is a normal, expected condition (lock held)
+            sys.exit(0)
```
dspachos, line 79: Not sure I understand this: The script intentionally return 0 (success) but how we track failures? 
Same for `scripts/trigger_hard_delete_job.py:78` , `scripts/trigger_sync_pool_budgets_job.py:80`

dspachos (approved):
Looks good to me:  Timezone semantics changed from explicit UTC to platform-default cron timezone (which even if it's not UTC it's ok).

One minor comment though in `scripts/trigger_recon_job.py:78`

scripts/trigger_recon_job.py
dan2k3k4: mmm that seems to come from
> Wrong exit code — sys.exit(1) was used when the lock was held, which would be reported as a cronjob failure.

which is Copilot change based on @alagoa's comment above

PR #376, opened 2026-04-01
chore: fix dedicated region keys

dan2k3k4:
- [x] Fix `create_vector_db`: move `force_user_keys` check before dedicated-region access check so the access check uses the effective owner/team after any rewrite
- [x] Fix `create_token`: compute `lite_team_id` once and reuse it in both the bootstrap call and `create_key`, eliminating the duplicate `format_team_id` computation
- [x] Syntax verified
- [x] Code review passed (no comments)
- [x] CodeQL scan passed (0 alerts)


dan2k3k4:
@dspachos I think you said this is not needed anymore?

PR #394, opened 2026-04-03
AI-291: Usage/spend per user

dspachos:
- [x] Explore codebase and understand current implementation
- [x] Precompute `user_id_strs` set once in `_fetch_region_spend` (avoid rebuild per-key)
- [x] Fix N+1 query: replace per-(team,region) DB queries with a single batch query
- [x] Consolidate email validation to use same constraints as `/users/by-email` (no regex, ReDoS-safe)
- [x] Redesign cache write to be atomic (`INSERT ... ON CONFLICT DO UPDATE`), removing `pg_advisory_xact_lock`
- [x] CodeQL 0 alerts confirmed


dspachos:
@copilot apply changes based on the comments in [this thread](https://github.com/amazeeio/amazee.ai/pull/394#pullrequestreview-4054503313) 
For the cache use the "redesign the cache write to be atomic..."

app/api/users.py
```diff
@@ -26,8 +38,14 @@
...
 from datetime import datetime, UTC
 import logging
+import asyncio
+import httpx
+from app.services.litellm import LiteLLMService
 
 logger = logging.getLogger(__name__)
+_USER_SPEND_CACHE_TTL_SECONDS = 15 * 60
+_USER_SPEND_TIMEOUT_SECONDS = 10.0
+_USER_SPEND_SEMAPHORE = asyncio.Semaphore(10)
```
dan2k3k4, line 48: Use env vars to allow changing these numbers, with defaulting to the values, e.g.:

assuming we already have:
```
import os
import asyncio
```

Then
```
# TTL - Defaults to 900 if not found
_USER_SPEND_CACHE_TTL_SECONDS = int(os.environ.get("USER_SPEND_CACHE_TTL_SECONDS", 15 * 60))

# Timeout - Defaults to 10.0
_USER_SPEND_TIMEOUT_SECONDS = float(os.environ.get("USER_SPEND_TIMEOUT_SECONDS", 10.0))

# Semaphore - Defaults to 10
_USER_SPEND_SEMAPHORE_LIMIT = int(os.environ.get("USER_SPEND_SEMAPHORE_LIMIT", 10))
_USER_SPEND_SEMAPHORE = asyncio.Semaphore(_USER_SPEND_SEMAPHORE_LIMIT)
```

tests/test_users.py
```diff
@@ -208,6 +212,270 @@ def test_create_user_by_team_admin(client, team_admin_token, test_team, db):
     assert any(u["id"] == user_data["id"] for u in team_data["users"])
 
 
+@patch("app.api.users.LiteLLMService.get_team_info", new_callable=AsyncMock)
```
dan2k3k4, line 215: I wonder if all the `@patch` in this test can be refactored into a single `@fixture` with one or two params to adjust the team / region / user info being passed in?

app/schemas/models.py
```diff
@@ -603,3 +603,24 @@ class PoolRegionPurchaseHistoryResponse(BaseModel):
     region_id: int
     purchases: List[PoolRegionPurchaseHistoryItem]
     model_config = ConfigDict(from_attributes=True)
+
+
+class UserSpendRegion(BaseModel):
+    region_id: int
+    region_name: str
```
dan2k3k4, line 610: Do we need to track the region name if it's already defined in the region model?

As the region name could be updated in the region model but not updated here?

I assume region_id can be used for JOIN queries

pmelab (approved):

PR #403, opened 2026-04-08
feat(auth): expiry and read/write scopes for management API tokens

dan2k3k4:
Rebased onto current `dev` (the branch had a stale `dev` merge that would have reverted `public.py`, `subscription.py` and `restore_database.py`) and squashed into two commits: expiry, then scopes.

## 1. Expiry

`api_tokens` gains `expires_at` and `expiry_option`. Existing tokens get `expiry_option='forever'` — which is what they already are. Expired tokens are rejected with 401 at auth time, and the token page gains an expiry selector.

The selectable lifetimes (1 day → forever) are a constant, `API_TOKEN_EXPIRY_OPTIONS` in `app/api/auth.py`, not the `api_token_expiry_options` table the earlier version of this PR added. The table needed seeding in three places to work — migration, `init_db`, and `initialise_resources` — because `initialise_resources.py` bootstraps a brand-new environment with `create_all` + `stamp head`, i.e. **without running migrations**, so a migration-only seed left fresh environments with zero rows and `POST /auth/token` returning 400 for every request, including the default `forever`. Nothing manages these rows at runtime, so a constant removes that whole failure mode.

`POST /auth/token` returns 400 for an unknown slug; `GET /auth/token/expiry-options` serves the list to authenticated callers.

## 2. Scopes

`api_tokens.scope` is `'read' | 'write'`, NOT NULL, `server_default 'read'`, validated by a `Literal` on `APITokenCreate`. **Read is the default** — write is an explicit opt-in. Scope is fixed at creation; re-create a token to change it. Anyone who can create a token can choose its scope; RBAC still bounds what the token can actually do.

Enforcement is one rule in `get_current_user_from_auth`: a read-scoped token may only use `GET`/`HEAD`/`OPTIONS`. I enumerated all 60 mutating routes in the API — every one is a genuine write, so the HTTP method is a reliable proxy and no route needs an annotation. An endpoint added later is covered on day one, rather than fail-open until someone remembers to decorate it.

The check runs in two places on purpose: `AuthMiddleware` resolves the token once per request and stores the user in `request.state`, and the endpoint's own `Depends(get_current_user_from_auth)` then returns from that state **without re-reading the token row** — so a check only at the resolution site would never run for a normal request. The middleware now passes the request through so the scope reaches `request.state`, and the dependency checks it before the short-circuit. JWT/cookie sessions carry no token scope and are unaffected.

### Backfill and rollout

The migration sets `scope='write'` for existing tokens owned by a system admin, and read for everything else:

- **Keeps working:** the admin-owned machine integrations — PolyDock (`aitrial-polydock-engine-*`, `private-gpt-backend-*`) and human admin tokens.
- **Breaks until moad ships:** moad's per-user `drupal_management_token` rows are owned by non-admin users and `POST /private-ai-keys`. They become read-only and will get 403 on writes until moad mints tokens with `scope: "write"`. **This needs a coordinated release with the moad repo.**

`tests/test_internal_provision_key.py` now creates moad's fixture token with `scope="write"`, which is the same change moad needs to make for real.

### Known gaps

Two side-effecting GETs stay reachable by a read token: `GET /billing/teams/{id}/pricing-table-session` (lazily creates and caches a Stripe customer id) and `GET /teams/sales` (caches computed metrics). Both are lazy-init caches rather than user-directed writes, so they are deliberately not special-cased.

## Review feedback from the previous rounds

Carried over: `is_active` non-null (moot — table gone), authenticated expiry-options endpoint, inactive-slug rejection (moot — no rows to deactivate), `logger.exception` instead of a silent `except`, guarded `downgrade()`, React Query as the single source of truth for the token list, empty-options fallback in the `<Select>`. Dropped as no longer applicable: the `_SessionProxy` seeding fixture and the `db`-fixture seeding call that Greptile flagged as redundant.

## Verification

- Full backend suite green on the rebased branch.
- Migration chain re-pointed to the current `dev` head (`e5b8c7d2a941`); single head confirmed.
- Both migrations exercised against PostgreSQL 16 on the real upgrade path (schema present, stamped at the parent revision): columns added, `expiry_option` defaults to `forever` and `scope` to `read` for rows inserted without them, admin-owned tokens backfilled to `write` and non-admin left `read`, and both `downgrade()`s clean up.
- Also set `server_default` on `expiry_option`/`scope` in the model, so a schema built by `create_all` matches the one built by migrations — they differed otherwise.

### DIFF
--- a/app/db/init_db.py
+++ b/app/db/init_db.py
@@ -7,7 +7,7 @@ logger = logging.getLogger(__name__)
 
 
 def init_api_token_expiry_options(db=None):
-    print("Initializing API token expiry options...")
+    logger.info("Initializing API token expiry options...")
     own_db = False
     if db is None:
         db = SessionLocal()
@@ -43,10 +43,11 @@ def init_api_token_expiry_options(db=None):
                 db.add(db_opt)
 
         db.commit()
-        print("API token expiry options initialized successfully!")
-    except Exception as e:
-        print(f"Error initializing API token expiry options: {e}")
+        logger.info("API token expiry options initialized successfully!")
+    except Exception:
+        logger.exception("Error initializing API token expiry options")
         db.rollback()
+        raise
     finally:
         if own_db:
             db.close()
--- a/app/db/init_db.py
+++ b/app/db/init_db.py
@@ -1,9 +1,13 @@
+import logging
+
 from app.db.models import Base, DBAPITokenExpiryOption
 from app.db.database import engine, SessionLocal
 
+logger = logging.getLogger(__name__)
+
 
 def init_api_token_expiry_options():
-    print("Initializing API token expiry options...")
+    logger.info("Initializing API token expiry options...")
     db = SessionLocal()
     try:
         options = [
@@ -35,18 +39,19 @@ def init_api_token_expiry_options():
                 db.add(db_opt)
 
         db.commit()
-        print("API token expiry options initialized successfully!")
+        logger.info("API token expiry options initialized successfully!")
     except Exception as e:
-        print(f"Error initializing API token expiry options: {e}")
+        logger.exception("Error initializing API token expiry options: %s", e)
         db.rollback()
+        raise
     finally:
         db.close()
 
 
 def init_db():
-    print("Creating database tables...")
+    logger.info("Creating database tables...")
     Base.metadata.create_all(bind=engine)
-    print("Database tables created successfully!")
+    logger.info("Database tables created successfully!")
     init_api_token_expiry_options()
 
 
--- a/app/db/init_db.py
+++ b/app/db/init_db.py
@@ -6,9 +6,13 @@ from app.db.database import engine, SessionLocal
 logger = logging.getLogger(__name__)
 
 
-def init_api_token_expiry_options():
-    logger.info("Initializing API token expiry options...")
-    db = SessionLocal()
+def init_api_token_expiry_options(db=None):
+    print("Initializing API token expiry options...")
+    own_db = False
+    if db is None:
+        db = SessionLocal()
+        own_db = True
+
     try:
         options = [
             {"name": "1 day", "slug": "1_day", "days": 1},
@@ -39,13 +43,13 @@ def init_api_token_expiry_options():
                 db.add(db_opt)
 
         db.commit()
-        logger.info("API token expiry options initialized successfully!")
+        print("API token expiry options initialized successfully!")
     except Exception as e:
-        logger.exception("Error initializing API token expiry options: %s", e)
+        print(f"Error initializing API token expiry options: {e}")
         db.rollback()
-        raise
     finally:
-        db.close()
+        if own_db:
+            db.close()
 
 
 def init_db():
--- a/app/db/init_db.py
+++ b/app/db/init_db.py
@@ -1,11 +1,53 @@
-from app.db.models import Base
-from app.db.database import engine
+from app.db.models import Base, DBAPITokenExpiryOption
+from app.db.database import engine, SessionLocal
+
+
+def init_api_token_expiry_options():
+    print("Initializing API token expiry options...")
+    db = SessionLocal()
+    try:
+        options = [
+            {"name": "1 day", "slug": "1_day", "days": 1},
+            {"name": "1 week", "slug": "1_week", "days": 7},
+            {"name": "1 month", "slug": "1_month", "days": 30},
+            {"name": "2 months", "slug": "2_months", "days": 60},
+            {"name": "3 months", "slug": "3_months", "days": 90},
+            {"name": "4 months", "slug": "4_months", "days": 120},
+            {"name": "5 months", "slug": "5_months", "days": 150},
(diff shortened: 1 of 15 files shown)


dan2k3k4:
@copilot apply changes based on the comments in [this thread](https://github.com/amazeeio/amazee.ai/pull/403#pullrequestreview-4078167178)

app/migrations/versions/20260408_125946_daf5bf0b03c2_add_api_token_expiry_options.py
```diff
@@ -0,0 +1,75 @@
...
+    op.create_table(
+        "api_token_expiry_options",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("name", sa.String(), nullable=False),
+        sa.Column("slug", sa.String(), nullable=False),
+        sa.Column("days", sa.Integer(), nullable=True),
+        sa.Column("is_active", sa.Boolean(), nullable=True, server_default="true"),
+        sa.PrimaryKeyConstraint("id"),
+    )
+    op.create_index(
```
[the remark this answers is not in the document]
dan2k3k4: @copilot apply changes based on [this feedback](https://github.com/amazeeio/amazee.ai/pull/403#discussion_r3054100338)

app/db/init_db.py
```diff
@@ -1,11 +1,62 @@
...
+                .filter(DBAPITokenExpiryOption.slug == opt_data["slug"])
+                .first()
+            )
+            if not existing:
+                db_opt = DBAPITokenExpiryOption(**opt_data)
+                db.add(db_opt)
+
+        db.commit()
+        print("API token expiry options initialized successfully!")
+    except Exception as e:
```
alagoa: **thought:** the PR's description mentioned we re-raise the exception here, but we are just printing the error and rolling back.

app/db/init_db.py
```diff
@@ -1,11 +1,62 @@
...
+import logging
+
+from app.db.models import Base, DBAPITokenExpiryOption
+from app.db.database import engine, SessionLocal
+
+logger = logging.getLogger(__name__)
+
+
+def init_api_token_expiry_options(db=None):
+    print("Initializing API token expiry options...")
```
alagoa: **suggestion:** update this so that it uses the logger defined above

alagoa (commented):
Overall looks good, nice work!

Just some comments and another question - do we want to have a cron to clear expired tokens after some days, or do we want to keep a log of them in the database?

dan2k3k4:
@copilot address the comments
- re-raise exception
- use logger defined in init_db.py

dspachos (commented):
- Why do we need this? Is it a product decision? 

Other than that:

#### 1. **Merge conflicts** — `mergeable: CONFLICTING`
The PR currently has merge conflicts with `dev`. These need to be resolved before merging.

#### 2. **`expiry_option` default on existing tokens** (Migration)
```python
sa.Column("expiry_option", sa.String(), nullable=False, server_default="forever"),
```
This is correct for new/existing rows, but existing tokens that were created before this PR won't have an `expires_at` value — they'll be `NULL`, meaning they never expire. This is the right default behavior, but it's worth documenting as a conscious decision.

#### 3. **`conftest.py` seeds expiry options globally**
```python
# conftest.py
init_api_token_expiry_options(db=db)
```
This seeds expiry options for *every* test. If some tests don't need them, this adds unnecessary overhead. Not a blocker, but worth noting.

#### 4. **Open question from @alagoa: cleanup of expired tokens**
> "do we want to have a cron to clear expired tokens after some days, or do we want to keep a log of them in the database?"

This is unanswered. Currently expired tokens remain in the DB indefinitely — they're just rejected at auth time. A follow-up PR might want to add a cleanup job or a "revoked" status.

#### 5. **No loading/error state for expiry options in frontend**
The `<Select>` renders the fallback `"forever"` item when options are empty/undefined, but there's no spinner or error message while the expiry options query is loading or has failed. If the API is slow/down, the user sees a dropdown with just "forever" and no indication something is wrong.

#### 6. **`test_api_token_expiry.py` `seed_expiry_options` fixture is redundant with `conftest.py`**
Both `conftest.py` (line 37) and `test_api_token_expiry.py` (line 33) seed expiry options. The `conftest.py` version calls `init_api_token_expiry_options(db=db)` directly, while the test file uses `_SessionProxy` + `monkeypatch`. The `autouse` fixture in `test_api_token_expiry.py` runs *after* the conftest seeding, so it's seeding twice. One of these should be removed.

#### 7. **Missing `init_api_token_expiry_options()` call on app startup**
The `init_db()` function calls `init_api_token_expiry_options()`, but if the app doesn't call `init_db()` on startup (e.g., in production where Alembic handles migrations), new expiry options added in future won't be auto-seeded. The migration handles the initial seed, but there's no runtime check. This is fine for now since the options are static, but something to keep in mind if options ever change.

dspachos (commented):
@dan2k3k4 please fix failing tests and also address `Greptile` comments.

dspachos (approved):
@dan2k3k4 please resolve conflicts

PR #420, opened 2026-04-20
feat(api): add max_budget handling for team member spending regions

dspachos:


dspachos:
The comment in lines 245:247 in `app/api/users.py` is defensive, and in current codebase it is valid. The underline potential issue is that there is no DB constraint enforcing “one base email per team” (this is handled by the MOAD app). We may need to discuss this and perhaps add some kind of constraint on this codebase as well (unless I'm missing something).

pmelab (approved):

PR #422, opened 2026-04-20
fix: invalidate user spend cache after member budget set/clear

pmelab:
## Summary

- `GET /users/spend` caches responses in `DBUserSpendCache` for 15 minutes
- `PUT /{region_id}/team/{team_id}/member/{user_id}/budget` and the corresponding `POST .../budget/clear` update the DB and LiteLLM but never touch the cache
- Result: callers (including the MOAD portal) see a stale `max_budget` for up to 15 minutes after a change, making the UI appear unresponsive

## Fix

Add `_invalidate_user_spend_cache(db, user.email)` and call it in both `update_team_member_budget` and `clear_team_member_budget` **before** `db.commit()`. This deletes the cache row for the affected user's normalized email so the next `GET /users/spend` recomputes from live data.

The normalization mirrors `_normalize_email_for_lookup` in `users.py` (strips `+` suffixes before `@`).

## Test plan

- [ ] Set a member budget → `GET /users/spend?email=...` immediately returns the new `max_budget`
- [ ] Clear a member budget → `GET /users/spend?email=...` immediately returns `max_budget: null`
- [ ] CI passes

### DIFF
--- a/app/api/spend.py
+++ b/app/api/spend.py
@@ -23,6 +23,7 @@ from app.db.models import (
     DBTeam,
     DBTeamRegion,
     DBUser,
+    DBUserSpendCache,
 )
 from app.schemas.limits import OwnerType, ResourceType
 from app.schemas.models import (
@@ -38,6 +39,25 @@ from app.services.litellm import LiteLLMService
 
 router = APIRouter(tags=["spend"])
 logger = logging.getLogger(__name__)
+
+
+def _invalidate_user_spend_cache(db: Session, email: str) -> None:
+    """Delete the cached spend response for *email* so the next read is fresh.
+
+    Called after any mutation that changes a user's ``max_budget`` or spend
+    cap, because the ``GET /users/spend`` endpoint caches responses for
+    ``_USER_SPEND_CACHE_TTL_SECONDS`` and does not auto-invalidate on writes.
+    """
+    parts = email.lower().rsplit("@", 1)
+    if len(parts) == 2:
+        normalized = f"{parts[0].split('+')[0]}@{parts[1]}"
+    else:
+        normalized = email.lower()
+    db.query(DBUserSpendCache).filter(
+        DBUserSpendCache.normalized_email == normalized
+    ).delete(synchronize_session=False)
+
+
 MONTHLY_BUDGET_DURATION = "1mo"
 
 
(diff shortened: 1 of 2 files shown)


app/api/spend.py
```diff
@@ -7,6 +7,7 @@
 
 from app.core.config import settings
 from app.core.limit_service import DEFAULT_MAX_SPEND, LimitService
+from app.api.users import _normalize_email_for_lookup
```
alagoa, line 10: **thought:** we should drop the "_" on this method's name, since we're importing it from another file. 

May cause some confusion if we ever want to change the original method, as devs will assume it's not used anywhere else.

app/api/spend.py
```diff
@@ -38,6 +40,22 @@
...
+
+    Called after any mutation that changes a user's ``max_budget`` or spend
+    cap, because the ``GET /users/spend`` endpoint caches responses for
+    15 minutes (``_USER_SPEND_CACHE_TTL_SECONDS`` in ``app/api/users.py``)
+    and does not auto-invalidate on writes.
+    """
+    normalized = _normalize_email_for_lookup(email)
+    db.query(DBUserSpendCache).filter(
+        DBUserSpendCache.normalized_email == normalized
+    ).delete(synchronize_session=False)
```
alagoa, line 56: **suggestion:** Add a comment that explains that the `synchronize_session=False` will not update the in-memory cache, so any subsequent fetches to `DBUserSpendCache` in the same request can result in outdated data.

This is fine here because we're not doing anything else with the `DBUserSpendCache` during this request, but the fact that this is done inside the `_invalidate_user_spend_cache` method can cause confusion in case we reuse it in the future.

dspachos (approved):

dspachos:
Current `origin/dev` already implements its core behavior (invalidate user spend cache when team-member budget is set/cleared).

I compared PR #422 against `origin/dev`: its two commits are not present as exact SHAs, but the functionality is already on `dev` via PR #434 (same fix + follow-ups).

**Related commits on `dev`:**
1. `99ffc9e` — `fix: invalidate user spend cache after member budget set/clear`
2. `e6f906c` — `Apply suggestion from @Copilot` (ordering toward atomic write flow)
3. `2c5f798` — `Apply suggestion from @Copilot` (commit placement update in member budget update path)
4. `7c61f35` — `Apply suggestion from @Copilot` (cache helper made non-committing; flush-based)
5. `26ac7b9` — `fix: commit cache invalidation atomically in clear_team_member_budget`
6. `0d5fe79` — Merge commit for PR #434 (`fix/member-budget-cache-invalidation`)

PR #432, opened 2026-04-22
feat(api): add model aliases management for teams in dedicated regions

dspachos:


dspachos:
@copilot apply changes based on the comments in [this thread](https://github.com/amazeeio/amazee.ai/pull/432#pullrequestreview-4154560816)

dspachos:
@copilot apply changes based on the comments in [this thread](https://github.com/amazeeio/amazee.ai/pull/432#pullrequestreview-4154560816)

app/services/litellm.py
```diff
@@ -565,6 +568,59 @@ async def update_team_budget(
...
+            raw_aliases = {}
+        if not isinstance(raw_aliases, dict):
+            return {}
+        return {
+            str(alias): str(model)
+            for alias, model in raw_aliases.items()
+            if alias is not None and model is not None
+        }
+
+    async def _get_team_model_aliases_from_team_list(
```
pmelab: Do we need this? We know which litellm versions we have, no?

pmelab (changes requested):

app/services/litellm.py
dspachos: It as compatibility support for LiteLLM versions where /team/info does not include aliases. But I believe all LiteLLM instances return aliases via `/team/info` so I guess it's safe to remove.

dspachos: Fixed

pmelab (approved):

PR #439, opened 2026-04-28
fix: skip per-key limits for pool budget teams

pmelab:
## Problem

Teams created through moad use `budget_type = pool`. Their spending ceiling is managed entirely at the team level via `purchase_pool_budget`, which sets the LiteLLM team-level `max_budget`. Per-key limits (`user_key: 1`, `service_key: 5`, `max_budget`, `rpm_limit`, `duration`) are meaningless in this model and actively block key creation once the default count limit is hit.

The user-count check already had a pool bypass (`check_team_user_limit` returns early for pool teams). Key limits had no equivalent bypass.

## Changes

**`app/api/private_ai_keys.py`**
- Resolve the team's `budget_type` before the limits block in `create_llm_token`
- Skip `check_key_limits` for pool teams (no count enforcement)
- Skip `get_token_restrictions` for pool teams — pass `None` for `days_left_in_period`, `max_max_spend`, `max_rpm_limit`
- Pass `apply_limits=False` to `LiteLLMService.create_key` for pool teams

**`app/services/litellm.py`**
- Add `apply_limits: bool = True` parameter to `create_key`
- Gate `budget_duration`, `max_budget`, `rpm_limit` on `apply_limits` (not just `ENABLE_LIMITS`)
- `duration` / `max_budget` / `rpm_limit` params become `Optional` to accommodate `None` from pool-team callers
- Key expiry (`duration = "365d"`) is always set regardless of `apply_limits` — pool team keys still expire

## Consistency with existing behaviour

| Check | Periodic teams | Pool teams (before) | Pool teams (after) |
|---|---|---|---|
| User count limit | enforced | **bypassed** (existing) | bypassed |
| Key count limit | enforced | enforced | **bypassed** |
| Per-key budget/RPM | enforced | enforced | **bypassed** |
| Key expiry | enforced | enforced | enforced |

### DIFF
--- a/app/api/private_ai_keys.py
+++ b/app/api/private_ai_keys.py
@@ -11,6 +11,7 @@ from app.schemas.models import (
     PrivateAIKeyCreate,
     PrivateAIKeySpend,
     BudgetPeriodUpdate,
+    BudgetType,
     LiteLLMToken,
     VectorDBCreate,
     VectorDB,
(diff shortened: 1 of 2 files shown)


dspachos:
I compared PR #439 (1160a79) against current dev, and the patch is already present there as commit   `3c77879` (fix: skip per-key limits for pool budget teams) touching the same two files with the same  change scope. dev also has a follow-up hardening commit `155fe48` (adds required-limit validation +  tests), so dev is already at least as complete as #439 for this behavior.

PR #494, opened 2026-05-14
AI-480 periodic team top up purchases

dspachos:

This PR introduces periodic team top-up purchases for PERIODIC billing teams, adding a FIFO budget ledger (`DBPeriodicBudgetLedgerEntry`) that tracks subscription allocations, top-ups, and rollovers, along with new `/subscription/cycle` and `/subscription/deactivate` MOAD API endpoints and a Stripe webhook idempotency layer (`DBStripeProcessedEvent`).

- **Ledger service** (`periodic_budget_ledger_service.py`): implements FIFO spend allocation, top-up rollover materialization, and subscription entry expiry; subscription budget is consumed before top-up budget, and remaining top-up balance rolls over at renewal.
- **New subscription endpoints** (`subscription.py`): `/cycle` triggers a billing-cycle pipeline (spend capture → ledger sync → LiteLLM budget update) authenticated with a MOAD API key; `/deactivate` zeros out a team's LiteLLM budget on subscription cancellation.
- **Webhook improvement** (`webhooks.py` + `worker.py`): `invoice.paid` events now drive the same `/cycle` pipeline directly, with claim-row idempotency to prevent duplicate processing of retried Stripe webhooks.
</details>

<details open><summary><h3>Confidence Score: 3/5</h3></summary>

Not safe to merge as-is: passing `None` instead of `payment_id` in `subscription.py` means any LiteLLM sync failure via the `/cycle` endpoint leaves the payment record permanently marked as succeeded, and a subsequent retry is silently dropped by the idempotency guard.

The `/cycle` endpoint passes `source_payment_id=None` to `apply_billing_cycle_for_team`, so when LiteLLM sync fails the payment record is never updated from its initial `success` state. On retry the idempotency check skips processing entirely. The `is_first_cycle` check also ignores `region_id` in both the `/cycle` endpoint and the webhook worker, causing spurious spend captures for multi-region periodic teams on a region's first billing cycle. Both defects are on the primary billing code path.

`app/api/subscription.py` has both critical issues; `app/core/worker.py` carries the same `is_first_cycle` scope problem; `app/db/models.py` and `app/core/periodic_budget_ledger_service.py` warrant a second look for the ineffective rollover uniqueness constraint.
</details>

| Filename | Overview |
|----------|----------|
| app/api/subscription.py | New MOAD-auth endpoints `/cycle` and `/deactivate` for periodic billing. Contains two bugs: `source_payment_id=None` instead of `payment_id` breaks sync-status tracking and makes failed LiteLLM syncs silently unretryable; `is_first_cycle` not scoped to `region_id` causes incorrect spend capture on first-cycle calls for multi-region teams. |
| app/core/periodic_budget_ledger_service.py | New FIFO ledger service for subscription/topup/rollover budget allocation. Core allocation logic looks correct; rollover expiry calculation can silently overwrite a `None` expiry with a finite date from another source entry. |
| app/core/worker.py | Major expansion adding `_run_cycle_from_stripe_event`, `_sync_periodic_ledger_for_period`, `apply_billing_cycle_for_team`. The `is_first_cycle` check in `_run_cycle_from_stripe_event` is not scoped to target regions, mirroring the same bug in `subscription.py`. |
| app/api/webhooks.py | New claim-row idempotency for incoming Stripe events via `DBStripeProcessedEvent`. Pattern is sound: duplicate webhooks are handled via `IntegrityError` before dispatching the background task. |
| app/db/models.py | Adds `DBPeriodicBudgetLedgerEntry` and `DBStripeProcessedEvent` models. The `uq_periodic_ledger_rollover_source_invoice_not_null` partial unique index is ineffective because `rolled_over_from_id` is always NULL and PostgreSQL unique indexes treat NULLs as distinct. |
| app/api/budgets.py | Adds `/purchase/periodic` top-up endpoint and `/periodic-status` status endpoint. Transaction handling is careful: payment record starts as `sync_failed` and is only updated to `success` after LiteLLM sync succeeds. |
| app/migrations/versions/20260515_120000_6e9b1c2d4f8a_periodic_ledger_partial_unique_indexes.py | Replaces full unique constraints with partial unique indexes (WHERE source_invoice_id IS NOT NULL). The rollover index on `[rolled_over_from_id, source_invoice_id]` provides no constraint when `rolled_over_from_id` is NULL. |

</details>

</details>

<details><summary><h3>Sequence Diagram</h3></summary>

```mermaid
sequenceDiagram
    participant Stripe
    participant Webhook as /billing/events
    participant BG as handle_stripe_event_background
    participant Cycle as _run_cycle_from_stripe_event
    participant MOAD as /subscription/cycle
    participant Ledger as periodic_budget_ledger_service
    participant LiteLLM

    Stripe->>Webhook: invoice.paid
    Webhook->>Webhook: Insert DBStripeProcessedEvent (idempotency)
    Webhook->>BG: background_tasks.add_task(event)
    Webhook-->>Stripe: 200 OK

    BG->>Cycle: _run_cycle_from_stripe_event(event)
    Cycle->>Cycle: Resolve team, region, budget_cents
    Cycle->>Cycle: Check DBPeriodicPayment idempotency
    Cycle->>Ledger: capture_periodic_team_spend_for_period
    Cycle->>Cycle: _record_periodic_payment_direct
    Cycle->>Ledger: _sync_periodic_ledger_for_period
    Ledger->>Ledger: allocate_period_spend_fifo
    Ledger->>Ledger: materialize_topup_rollovers
    Ledger->>Ledger: expire_subscription_entries
    Ledger->>Ledger: add_subscription_entry
    Cycle->>LiteLLM: apply_billing_cycle_for_team
    LiteLLM-->>Cycle: ok / error
    Cycle->>Cycle: Update sync_status, db.commit

    MOAD->>MOAD: Check DBPeriodicPayment idempotency
    MOAD->>Ledger: Same pipeline as Cycle above
    MOAD->>LiteLLM: apply_billing_cycle_for_team
    LiteLLM-->>MOAD: ok / error
    MOAD-->>MOAD: SubscriptionCycleResponse
```
</details>

<sub>Reviews (1): Last reviewed commit: ["feat(api): implement idempotency for Str..."](https://github.com/amazeeio/amazee.ai/commit/9c680075cd4a1be30380eed076fbbcc90b916d83) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34160171)</sub>

> Greptile also left **5 inline comments** on this PR.


app/migrations/versions/20260513_190000_b7e1c2d3f4a5_periodic_budget_ledger_and_stripe_event_idempotency.py
```diff
@@ -0,0 +1,190 @@
...
+
+revision = "b7e1c2d3f4a5"
+down_revision = "2f7c9d1e4aab"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "stripe_processed_events",
```
pmelab, line 20: **`stripe_processed_events` table appears to be orphaned.**

This migration (and `f0a1b2c3d4e5` adding a `created_at` index) creates the table with a unique constraint on `stripe_event_id`, but I can find no code in `app/` or `tests/` that references the table — no ORM model in `app/db/models.py`, no reads/writes in `app/api/webhooks.py` or `app/core/worker.py`.

The current webhook idempotency is enforced late, via `DBPeriodicPayment.stripe_payment_id` inside `_run_cycle_from_stripe_event` (worker.py ~514), after team/region/budget resolution — and only for periodic teams. Two concurrent deliveries of the same `invoice.paid` event can both pass that check before either hits the DB.

Was the intent to wire this table into `webhooks.py` as a pre-processing claim row (insert before scheduling the background task, catch `UniqueViolation` = already claimed)? If so, the wiring is missing. If not, the migration should be dropped.

Not blocking — just want to confirm intent.

app/core/worker.py
```diff
@@ -421,199 +763,431 @@ async def capture_periodic_team_spend_for_invoice(
...
+                "Skipping billing cycle sync for team %s: region %s not found",
+                team.id,
+                region_id,
             )
+            return sync_errors
 
-            # Update the team-level budget (shared ceiling for all keys)
-            lite_team_id = LiteLLMService.format_team_id(region.name, team.id)
+        team.last_payment = period_start
+        db.commit()
```
pmelab: **`team.last_payment` is committed before LiteLLM sync runs.** Two concerns:

1. **Failed sync still bumps `last_payment`.** If anything in lines 1079–1146 raises, `last_payment` was already committed, so retention/freshness metrics will show the team as "recently paid" despite a failed cycle.
2. **Early commit closes the transaction.** The subsequent `allocate_period_spend_fifo` (via `_active_entries`) uses `SELECT … FOR UPDATE`. With the commit at line 1071, those row locks are acquired in a fresh implicit transaction and don't span the rest of the cycle's work — defeating their purpose.

Suggestion: move the assignment + commit to after the LiteLLM sync succeeds, inside the same transaction as the ledger updates.

Separately: all three callers (`_run_cycle_from_stripe_event` worker.py:525, `/billing/subscription/cycle` subscription.py:128, and `apply_product_for_team`) pass `datetime.now(UTC)` as `period_start`, not the real Stripe invoice `period_start`. Worth plumbing the actual invoice timestamp through so `last_payment` reflects Stripe's truth, not the moment the webhook landed.

pmelab (commented):
Three items from triage. Two inline, one summary.

- [x] **README env vars not updated.** The Environment Variables section in README.md (lines ~285–296) doesn't reflect this PR. Please add `MOAD_API_KEY` and `PERIODIC_TOPUP_EXPIRY_DAYS`, and remove `WEBHOOK_SIG`. Endpoint-level detail is already covered in `docs/workflows.md` Section 5, so no further README changes needed beyond env vars.

app/api/subscription.py
```diff
@@ -0,0 +1,371 @@
...
+        )
+
+    region = db.query(DBRegion).filter(DBRegion.id == request.region_id).first()
+    if not region:
+        raise HTTPException(
+            status_code=404, detail=f"Region {request.region_id} not found"
+        )
+
+    period_start = datetime.now(UTC)
+    period_end = period_start + timedelta(days=31)
```
pmelab, line 132: **Drop the hardcoded 31d and let LiteLLM stop running its own clock.**

The `+31d` here and `budget_duration="31d"` pushed to LiteLLM in `apply_billing_cycle_for_team` (worker.py:1076,1113) form a parallel-clock pair that drifts as soon as `/cycle` doesn't land exactly every 31 days:

- **Faster than 31d** → LiteLLM resets too often + ledger expires the still-young sub on each call → silent over-allocation.
- **Slower than 31d** → LiteLLM self-resets at day 31 and grants fresh budget for the gap → silent under-billing.

With MOAD as the source of truth for cycle boundaries, the platform should own all reset timing via explicit `update_team_budget` calls from `/cycle`. Stop pushing `budget_duration` to LiteLLM (push `None`, or whatever LiteLLM accepts as "never auto-reset") and let `team.spend` reset only when `/cycle` lands. Missed cycle → team stays locked at the cap until the next `/cycle` arrives — preferable to either drift direction.

Audit found three 31d literals to remove: `app/api/budgets.py:55` (`PERIODIC_BUDGET_DURATION`), `app/core/worker.py:1076` and `:1294`, and this line. The other places that touch the duration — `expire_subscription_entries(period_end)`, `materialize_topup_rollovers(rollover_at)`, `add_subscription_entry(expires_at)` — are cosmetic or load-bearing only for the `/periodic-status` UI.

app/api/spend.py
```diff
@@ -842,6 +902,45 @@ async def get_team_spend(
...
+            .all()
+        )
+        for row in sub_rows:
+            sub_remaining_cents += max(0, row.amount_cents - row.consumed_cents)
+        topup_remaining_cents = compute_active_topup_remaining(
+            db, team_id=team_id, region_id=region_id
+        )
+        purchased_cents = sub_remaining_cents + topup_remaining_cents
+        purchased_dollars = purchased_cents / 100.0
+        live_remaining_dollars = max(0.0, purchased_dollars - float(total_spend or 0.0))
```
pmelab, line 941: **Lock the "consumed_cents only changes at cycle close" invariant down with a test.**

This formula looks like double-subtraction at first read: `purchased_cents` is already `sum(amount_cents - consumed_cents)` across active ledger rows, and `total_spend` is the same spend that has presumably been allocated against the ledger. It happens to be correct *today* because of a subtle, undocumented invariant — `consumed_cents` is only incremented at cycle close (by `allocate_period_spend_fifo` in `_sync_periodic_ledger_for_period`), and LiteLLM's `total_spend` is simultaneously reset on the same call, so the two never both reflect the same dollars at the same time.

If a future change ever allocates spend into the ledger mid-cycle (e.g. a top-up purchase that runs FIFO eagerly), `live_remaining` would silently under-report by double-counting that spend.

Add a unit test that asserts `periodic_budget.remaining_budget_cents` matches `compute_active_topup_remaining + sub_remaining` across a few `consumed_cents`/`total_spend` configurations, including "right after cycle close" and "mid-cycle with traffic". Also worth a docstring on the formula explaining the invariant.

app/services/litellm.py
```diff
@@ -448,6 +448,52 @@ async def get_team_info(self, team_id: str) -> dict:
                 detail=f"Failed to get LiteLLM team info: {error_msg}",
             )
 
+    async def get_spend_logs(
```
pmelab: **Dead code in this PR — no callers anywhere in `app/` or `tests/`.**

`grep -rn "get_spend_logs"` finds zero callers. The actual snapshot helpers (`capture_periodic_team_spend_for_period`, etc.) use `get_team_info`, not this.

Either wire up the intended caller and add a test, or drop the method from this PR and add it together with its first user in a follow-up. Shipping unused public API also locks in design choices (the `master_key` auth, the 1-500 limit clamp, the empty-list fallback for non-list responses, raising `HTTPException` from a service-layer helper) that haven't been validated against real usage yet.

pmelab (commented):
Three points captured during a code walkthrough — bundled as inline comments below. Non-blocking.

pmelab (approved):

pmelab (approved):

PR #527, opened 2026-05-22
chore: add restore database script

dan2k3k4:
<sub>Reviews (5): Last reviewed commit: ["chore: address restore script PR feedbac..."](https://github.com/amazeeio/amazee.ai/commit/e8782efe7c1f0960652e9a42e6911411d370fc7e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33402995)</sub>

Adds a new `scripts/restore_database.py` that restores a PostgreSQL database from a Lagoon `.tar.gz` backup via `pg_restore --create --clean`. The script uses `pg_dump -Fc` for a pre-restore safety backup, strips the password from URLs before they reach `argv`, and validates tar archive paths to prevent traversal.

- Password-stripping logic in `_strip_password_from_netloc` ensures credentials flow only through `PGPASSWORD`, never via `--dbname` on the command line.
- `safe_extract_tar` rejects symlinks, hard links, and path-traversal members, and passes a pre-validated member list to `extractall` to close the double-traversal race noted in earlier review rounds.
- Backup failure in non-interactive (`--yes`) mode now correctly aborts with `sys.exit(1)` rather than silently proceeding to the drop step.
</details>

Safe to merge — all critical issues identified in previous review rounds have been addressed, and no new defects were found.

The script now correctly strips the database password before it reaches argv, uses pg_dump -Fc for a structurally complete safety backup, blocks symlinks and path-traversal entries in the tar archive, and aborts rather than silently proceeding when a backup fails in non-interactive mode. The remaining findings are cosmetic or concern future Python compatibility.

scripts/restore_database.py — the only changed file; no other files require attention.
</details>

| Filename | Overview |
|----------|----------|
| scripts/restore_database.py | New restore script. All critical issues from prior review rounds appear addressed. Two minor findings: psql is declared as required but never invoked, and tarfile.extractall() omits the filter parameter that Python 3.12+ expects. |

</details>

</details>

<sub>Reviews (6): Last reviewed commit: ["chore: safe url"](https://github.com/amazeeio/amazee.ai/commit/3822adec45feb7e48431d269675364afeeca2780) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33402995)</sub>


dspachos (commented):
#### 1. Safety backup is completely broken — binary/text mode mismatch

(throughout backup_current_database)

 The file is opened in binary mode ("wb") but every f.write(...) call passes a plain Python string. This raises TypeError: a bytes-like object is required, not
 'str' on the very first write, meaning no backup data is ever written. Since this is the only safety net before a destructive DROP DATABASE, a failed restore
 would leave the database unrecoverable.

 ```python
   # Current (broken):
   with os.fdopen(fd, "wb") as f:
       f.write("-- Pre-restore safety backup\n")  # TypeError!

   # Fix: encode all string writes, or open in text mode and use a separate binary file for COPY
 ```

 Either:
 - Encode all string writes: f.write("...".encode()), or
 - Open in text mode and handle copy_expert via a separate binary buffer

 #### 2. ALTER OWNER is a dead pattern — causes false-fatal aborts

 Line 468

 ```python
   non_critical_prefixes = ("COMMENT ", "GRANT ", "REVOKE ", "SET ", "SELECT ", "ALTER OWNER ")
 ```

 No PostgreSQL statement starts with "ALTER OWNER". pg_dump generates ownership transfers as ALTER TABLE ... OWNER TO ... or ALTER SEQUENCE ... OWNER TO ....
 When these fail (common when the target role doesn't exist), the script treats them as fatal and aborts — leaving a dropped, empty database. The latest commit
 appears to have fixed this with an in-body check (" OWNER TO " in block_upper), but the old dead prefix is still in the prefix tuple. Verify the latest commit
 actually removed it.

dspachos (commented):
### 1. `pg_restore` error diagnostics silently lost
**`restore_database.py:156`** — `apply_restore()`

```python
run_pg_tool(cmd, config, capture=False)
```

When `capture=False`, `subprocess.run` does **not** capture stdout/stderr — they stream to the terminal. But then in `run_pg_tool` (line 91), `result.stderr` is `None`, so the `CalledProcessError` is raised with an **empty** stderr string. The `apply_restore` SystemExit message becomes:

> `pg_restore failed: CalledProcessError(...)` — with **no actual error detail**

This is a destructive operation (database already dropped). If restore fails, the operator has **zero diagnostic information** about why.

**Fix:** Either always capture stderr for error handling, or capture separately:
```python
def run_pg_tool(cmd, config, *, capture=True, check=True):
    result = subprocess.run(
        cmd,
        env=pg_env(config),
        stdout=subprocess.PIPE if capture else None,
        stderr=subprocess.PIPE,        # ALWAYS capture stderr for sanitization
        text=True,
        check=False,
    )
```

### 2. Unsanitized `pg_restore` stderr leaks password to terminal
**`restore_database.py:156`** — `apply_restore()`

Related to #1: `capture=False` sends `pg_restore` stderr directly to the terminal **bypassing the `sanitize()` function**. If `pg_restore` includes connection info in error messages, the database password is logged in cleartext.

**Fix:** Same as above — always capture stderr, sanitize it, then print it yourself.

scripts/restore_database.py
```diff
@@ -0,0 +1,444 @@
...
+    """Remove sensitive data (password) from messages."""
+    if not msg:
+        return msg
+    pw = config.get("password")
+    if pw:
+        msg = msg.replace(pw, "****")
+    return msg
+
+
+def get_db_config():
```
dspachos, line 93: @dan2k3k4 can you check this? 
`get_db_config()` only keeps host/port/user/password/dbname, so any libpq params in the URL (sslmode, connect_timeout, socket host=..., etc.) are silently lost. That makes the restore path diverge from the configured connection and can break on SSL-only or socket-based Postgres setups.

scripts/restore_database.py
```diff
@@ -0,0 +1,444 @@
...
+            # Best-effort cleanup only: preserve original pg_dump failure path.
+            print(f"  Warning: could not remove partial backup file: {cleanup_err}", file=sys.stderr)
+        raise SystemExit(f"  pg_dump failed: {e.stderr or e}")
+
+    size_mb = os.path.getsize(backup_path) / (1024 * 1024)
+    print(f"  Backup complete ({size_mb:.1f} MB)")
+    return backup_path
+
+
+def drop_and_recreate_database(config):
```
dspachos: @dan2k3k4 also check this please: 
`createdb db_name` rebuilds with cluster defaults and the current role as owner, so a non-default source DB can come back with the wrong owner/encoding/collation/tablespace. For a restore tool, that’s a correctness gap.

dspachos (approved):
See my 2 comments please.

PR #528, opened 2026-05-26
AI-408 hard delete soft deleted teams

dspachos:

This PR implements the final stage of the team lifecycle: a daily background job that hard-deletes teams whose `deleted_at` is ≥ 90 days old (raised from 60), with a full FK-safe cascade covering `spend_caps`, `api_tokens`, `user_admin_regions`, audit-log nulling, and `user_spend_cache`. It also gates all API access for users whose team is soft-deleted (403), re-provisions LiteLLM on team restore, and adds audit-log entries for all three lifecycle events.

- **Hard-delete cascade** (`worker.py`): raises threshold to 90 days; adds deletion of `spend_caps`, `api_tokens`, `user_admin_regions`, `user_spend_cache`, and audit-log user_id nulling before deleting users/team; writes a `team.hard_delete` audit entry.
- **Suspension gate** (`security.py`): `_check_user_team_not_suspended` added to every auth path; users of soft-deleted teams receive 403 instead of reaching endpoint logic.
- **Restore re-provisioning** (`team_service.py`): `restore_soft_deleted_team` now re-creates the LiteLLM team and users in each active region and returns a `litellm_warnings` dict; the endpoint surfaces failures as a `warning` field.
</details>

<details open><summary><h3>Confidence Score: 3/5</h3></summary>

The hard-delete and auth-gate logic is well-structured, but an email-case mismatch in the spend-cache cleanup means PII may survive a GDPR erasure, and two auth-path edge cases in security.py could cause unexpected errors under specific conditions.

The spend-cache cleanup queries `normalized_email` (lowercase) using raw `DBUser.email` values that can be mixed-case, so user emails may remain in the cache after a hard delete. In security.py, the API-token path's joinedload is negated by an intervening commit, and the non-dict middleware branch now accesses a relationship on a potentially-detached object without any guard. These issues are concentrated in the two most critical files — the GDPR cleanup worker and the authentication layer.

app/core/worker.py (spend-cache email normalization) and app/core/security.py (API-token joinedload order, non-dict middleware branch) warrant a closer look before merging.
</details>

| Filename | Overview |
|----------|----------|
| app/core/worker.py | Hard-delete threshold raised from 60→90 days; adds cleanup of spend_caps, api_tokens, user_admin_regions, audit_log nulling, and user_spend_cache — but email case mismatch in spend_cache cleanup may leave PII after erasure. |
| app/core/security.py | Adds _check_user_team_not_suspended guard across all auth paths with joinedload for team; joinedload is wasted in the API-token path due to commit before access, and the non-dict middleware path may hit DetachedInstanceError. |
| app/core/team_service.py | restore_soft_deleted_team now re-provisions LiteLLM team/users per region and returns litellm_warnings dict; soft_delete adds audit log; logic is sound but continues user provisioning unnecessarily when team creation fails. |
| app/api/teams.py | restore_team endpoint now surfaces litellm_warnings from the service layer as a response warning field; clean and straightforward change. |
| tests/test_hard_delete.py | New tests cover api_tokens, user_admin_regions, audit log nulling, spend_caps, key spend_caps, and user_spend_cache cleanup; all use lowercase test emails so the email-case mismatch bug is not caught. |
| tests/test_team_retention.py | Updated to reflect 403 (suspended) instead of 404 for soft-deleted team access; adds audit log, re-provisioning, and orphan-detection tests; all correct. |
| Makefile | Refactors test lifecycle: separates test-teardown target, removes port-mapping from postgres container, adds teardown after each test run. |
| README.md | Adds Team Lifecycle & Hard Delete documentation section explaining the three-stage lifecycle, cascade order, restore behaviour, and manual trigger. |

</details>

</details>

<details><summary><h3>Flowchart</h3></summary>

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
    A[Cron: daily 03:00] --> B[hard_delete_expired_teams]
    B --> C{teams with deleted_at >= 90 days?}
    C -- No --> Z[Done]
    C -- Yes --> D[For each eligible team]
    D --> E[1. Delete limited_resources]
    E --> F[2. Delete LiteLLM keys per region]
    F --> G[3. Delete spend_caps]
    G --> H[4. Delete ai_tokens from DB]
    H --> I[5. Delete api_tokens + user_admin_regions]
    I --> J[6. NULL audit_logs.user_id]
    J --> K[7. Delete user_spend_cache]
    K --> L[8. Delete users]
    L --> M[9. Delete team_products + team_regions]
    M --> N[10. Write audit log team.hard_delete]
    N --> O[11. Delete team - cascades team_metrics]
    O --> P[COMMIT]
    P --> D

    subgraph Auth Gate
    Q[Incoming request] --> R[load user + team via joinedload]
    R --> S{team.deleted_at set AND not admin?}
    S -- Yes --> T[403 Suspended]
    S -- No --> U[Allow request]
    end
```
</details>

<sub>Reviews (1): Last reviewed commit: ["docs: Update README with team lifecycle ..."](https://github.com/amazeeio/amazee.ai/commit/ff68e78349eaa43a66fb4ed6c196d6eac8d30b59) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33938937)</sub>

> Greptile also left **4 inline comments** on this PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: backend-test backend-test-build test-clean test-network test-postgres frontend-test frontend-test-build migration-create migration-upgrade migration-downgrade migration-stamp
+.PHONY: backend-test backend-test-build test-clean test-teardown test-network test-postgres frontend-test frontend-test-build migration-create migration-upgrade migration-downgrade migration-stamp
 
 # Default target
 all: backend-test
@@ -12,20 +12,26 @@ backend-test-build:
 	docker build -t amazee-backend-test -f Dockerfile.test .
 
 # Start PostgreSQL container for testing
-test-postgres: test-clean test-network
+test-postgres: test-network
 	docker run -d \
 		--name amazee-test-postgres \
 		--network amazeeai_default \
 		-e POSTGRES_USER=postgres \
 		-e POSTGRES_PASSWORD=postgres \
 		-e POSTGRES_DB=postgres_service \
-		-p 5432:5432 \
 		pgvector/pgvector:pg16 && \
 	sleep 5
 
+# Teardown: stop and remove test containers, network, and image
+test-teardown:
+	docker stop amazee-test-postgres 2>/dev/null || true
+	docker rm amazee-test-postgres 2>/dev/null || true
+	docker network rm amazeeai_default 2>/dev/null || true
+	docker rmi amazee-backend-test 2>/dev/null || true
+
 # Run backend tests for a specific regex
 # Usage: make backend-test-regex regex="test_pattern"
-backend-test-regex: test-clean backend-test-build test-postgres
+backend-test-regex: backend-test-build test-postgres
 	@if [ -z "$(regex)" ]; then \
 		echo "Error: regex parameter is required. Usage: make backend-test-regex regex=\"test_pattern\""; \
 		exit 1; \
@@ -45,9 +51,10 @@ backend-test-regex: test-clean backend-test-build test-postgres
 		-v $(PWD)/app:/app/app \
 		-v $(PWD)/tests:/app/tests \
 		amazee-backend-test pytest -vv -k "$(regex)"
+	$(MAKE) test-teardown
 
 # Run backend tests in a new container
-backend-test: test-clean backend-test-build test-postgres
+backend-test: backend-test-build test-postgres
 	docker run --rm \
 		--network amazeeai_default \
 		-e DATABASE_URL="postgresql://postgres:postgres@amazee-test-postgres/postgres_service" \
@@ -63,9 +70,10 @@ backend-test: test-clean backend-test-build test-postgres
 		-v $(PWD)/app:/app/app \
 		-v $(PWD)/tests:/app/tests \
 		amazee-backend-test
+	$(MAKE) test-teardown
 
 # Run backend tests with coverage report
-backend-test-cov: test-clean backend-test-build test-postgres
+backend-test-cov: backend-test-build test-postgres
 	docker run --rm \
 		--network amazeeai_default \
 		-e DATABASE_URL="postgresql://postgres:postgres@amazee-test-postgres/postgres_service" \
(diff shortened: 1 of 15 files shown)


app/core/worker.py
```diff
@@ -1515,16 +1521,16 @@ async def monitor_teams(db: Session):
...
-    Hard deletion job for teams that have been soft-deleted for 60+ days.
+    Hard deletion job for teams that have been soft-deleted for 90+ days.
     Cascades deletion to all related resources (keys, users, limits, metrics, etc.).
     Runs less frequently than monitor_teams (daily at 3 AM).
     """
     logger.info("Starting hard delete job for expired teams")
     try:
-        # Calculate cutoff date (60 days ago)
-        cutoff_date = datetime.now(UTC) - timedelta(days=60)
+        # Calculate cutoff date (90 days ago — GDPR data retention requirement)
```
dan2k3k4: it is not specifically a GDPR data retention requirement... actually GDPR should be "store only as long as needed" so the shorter the better, but we likely may need to store/keep soft-deleted teams around for internal reasons

dan2k3k4: I would just drop/remove the comment, and perhaps expose the days as a variable so that we can set it / change it system wide if needed (maybe with a minimum amount e.g. `max(30, SYSTEM_RETENTION_DAYS)` [so it will be at least 30 days else the env var if it's bigger]

Makefile
```diff
@@ -12,24 +12,31 @@ backend-test-build:
...
 		--network amazeeai_default \
 		-e POSTGRES_USER=postgres \
 		-e POSTGRES_PASSWORD=postgres \
 		-e POSTGRES_DB=postgres_service \
-		-p 5432:5432 \
 		pgvector/pgvector:pg16 && \
 	sleep 5
 
+# Teardown: stop and remove test containers, network, and image
+test-teardown:
```
dan2k3k4: Doesn't `test-clean` do this (partially)?

Wouldn't this make tests slower as it has to spin up new docker containers each time instead of just cleaning the data inside the docker containers?

dspachos: Yes, actually that was not related to this task, but helps testing locally. Fixed it.

app/core/worker.py
dspachos: Fixed

Makefile
dspachos: Actually, `test-teardown` cleans leftovers from the previous test (if any) while `test-clean` vanishes docker stuff at the end of the current test (but if something goes wrong, leftovers main remain). Nevertheless, I removed.

dan2k3k4 (approved):

PR #534, opened 2026-05-29
chore: fix load spend button

dan2k3k4:

This PR fixes the "Load Spend" button for team-based AI keys by mapping the team spend API response shape (`total_spend`, `total_budget`) to the `SpendInfo` interface fields (`spend`, `max_budget`) expected by the frontend, and makes `expires`, `created_at`, and `updated_at` optional in the shared type since the team endpoint does not return them.

- The mapping (`total_spend ?? 0` → `spend`, `total_budget` → `max_budget`) is applied consistently in all three call sites: `team-expansion-row.tsx`, `user-expansion-row.tsx`, and `private-ai-key-spend-cell.tsx`.
- `SpendInfo` in `@/types/spend.ts` correctly marks the three per-key-only fields as optional to accommodate both the team and per-key spend endpoints.
</details>

The core fix is correct and consistent across all three call sites; one local interface copy in `private-ai-key-spend-cell.tsx` was not updated alongside the shared type.

The mapping logic (`total_spend ?? 0`, `total_budget`) is applied identically in all three affected files and the runtime display behaviour should work correctly. The only issue is that the local `SpendInfo` interface in `private-ai-key-spend-cell.tsx` still marks `expires`, `created_at`, and `updated_at` as required non-optional strings — the exact fields that were made optional in the shared type this PR updates. In a TypeScript strict build this mismatch between the local interface and the team-endpoint return value could surface as a compile-time error.

frontend/src/components/private-ai-key-spend-cell.tsx — its local SpendInfo interface was not updated to match the shared type change.
</details>

| Filename | Overview |
|----------|----------|
| frontend/src/types/spend.ts | Makes `expires`, `created_at`, and `updated_at` optional to match the team spend endpoint's response shape — a correct, minimal change. |
| frontend/src/app/admin/teams/_components/team-expansion-row.tsx | Maps `total_spend`/`total_budget` from the team spend response to `spend`/`max_budget` before storing in `spendMap`; uses the shared `SpendInfo` import correctly. |
| frontend/src/app/admin/users/_components/user-expansion-row.tsx | Applies the same `total_spend`/`total_budget` mapping as the team row; uses the shared `SpendInfo` import correctly. |
| frontend/src/components/private-ai-key-spend-cell.tsx | Applies the correct team spend mapping, but the local `SpendInfo` interface still marks `expires`, `created_at`, and `updated_at` as required strings — inconsistent with the shared type update and with what the team endpoint actually returns. |

</details>

</details>

<details><summary><h3>Sequence Diagram</h3></summary>

```mermaid
sequenceDiagram
    participant UI as Frontend Component
    participant API as Backend API
    UI->>API: "GET /spend/{regionId}/team/{teamId}"
    API-->>UI: "{ total_spend, total_budget, ... }"
    Note over UI: Map response:<br/>spend = total_spend ?? 0<br/>max_budget = total_budget
    UI->>UI: Render spend / budget display
    UI->>API: "GET /private-ai-keys/{keyId}/spend (fallback, no team)"
    API-->>UI: "{ spend, max_budget, expires, created_at, updated_at, ... }"
    UI->>UI: Use response directly (no mapping needed)
```
</details>

<details><summary><h3>Comments Outside Diff (1)</h3></summary>

1. `frontend/src/components/private-ai-key-spend-cell.tsx`, line 16-24 ([link](https://github.com/amazeeio/amazee.ai/blob/c9d4f25aa63a587cebdce1a9ad7b10c3e5569a02/frontend/src/components/private-ai-key-spend-cell.tsx#L16-L24)) 

   <a href="#">P1</a> The local `SpendInfo` interface still declares `expires`, `created_at`, and `updated_at` as required `string` fields, while the shared type in `@/types/spend.ts` was updated in this same PR to make them optional. When the team spend path returns a response that doesn't include these fields, the object returned by `queryFn` won't satisfy the local interface, which can produce a TypeScript error in strict mode. The component could simply drop the local definition and import the updated shared type instead.

</details>

<sub>Reviews (1): Last reviewed commit: ["chore: fix load spend button"](https://github.com/amazeeio/amazee.ai/commit/c9d4f25aa63a587cebdce1a9ad7b10c3e5569a02) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=34553894)</sub>

### DIFF
--- a/frontend/src/app/admin/teams/_components/team-expansion-row.tsx
+++ b/frontend/src/app/admin/teams/_components/team-expansion-row.tsx
@@ -35,6 +35,7 @@ import { SpendInfo } from "@/types/spend";
 import { Team } from "@/types/team";
 import { User } from "@/types/user";
 import { get } from "@/utils/api";
+import { mapTeamSpendToSpendInfo } from "@/utils/spend-mapping";
 import { useQuery } from "@tanstack/react-query";
 
 interface TeamExpansionRowProps {
@@ -159,12 +160,7 @@ export function TeamExpansionRow({
               `/spend/${matchedRegion.id}/team/${teamId}`,
             );
             const data = await response.json();
-            // Map TeamSpendResponse to SpendInfo
-            const spendInfo = {
-              ...data,
-              spend: data.total_spend ?? 0,
-              max_budget: data.total_budget,
-            };
+            const spendInfo = mapTeamSpendToSpendInfo(data);
             for (const keyId of keyIds) {
               spendData[keyId.toString()] = spendInfo;
             }
--- a/frontend/src/app/admin/teams/_components/team-expansion-row.tsx
+++ b/frontend/src/app/admin/teams/_components/team-expansion-row.tsx
@@ -158,7 +158,13 @@ export function TeamExpansionRow({
             const response = await get(
               `/spend/${matchedRegion.id}/team/${teamId}`,
             );
-            const spendInfo = await response.json();
+            const data = await response.json();
+            // Map TeamSpendResponse to SpendInfo
+            const spendInfo = {
+              ...data,
+              spend: data.total_spend ?? 0,
+              max_budget: data.total_budget,
+            };
             for (const keyId of keyIds) {
               spendData[keyId.toString()] = spendInfo;
             }
--- a/frontend/src/app/admin/users/_components/user-expansion-row.tsx
+++ b/frontend/src/app/admin/users/_components/user-expansion-row.tsx
@@ -15,6 +15,7 @@ import { PrivateAIKey } from "@/types/private-ai-key";
 import { Region } from "@/types/region";
 import { SpendInfo } from "@/types/spend";
 import { get, post } from "@/utils/api";
+import { mapTeamSpendToSpendInfo } from "@/utils/spend-mapping";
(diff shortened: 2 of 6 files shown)


dspachos (commented):
@dan2k3k4 can you take a look on the following? 

 ####  `use-private-ai-keys-data.ts` — Missing the same mapping (4th call site not patched)

 File: frontend/src/hooks/use-private-ai-keys-data.ts:89
 Problem: This hook also calls spend/${matchedRegion.id}/team/${teamId} and casts the response directly to SpendInfo without mapping total_spend → spend /
 total_budget → max_budget. If this code path is reachable, the "Load Spend" button will still be broken here. The regions variable is also referenced inside the
 query before it's defined (line 80 uses regions from the query declared below at line 97).

 ```ts
   // Line 89 — no mapping applied!
   const spendInfo: SpendInfo = await response.json();
   for (const keyId of keyIds) {
     result[keyId] = spendInfo;  // spendInfo.spend will be undefined
   }
 ```

 Fix: Apply the same mapping, or better — extract a shared helper:

 ```ts
   const raw = await response.json();
   const spendInfo: SpendInfo = {
     ...raw,
     spend: raw.total_spend ?? 0,
     max_budget: raw.total_budget ?? null,
   };

dspachos (approved):

PR #609, opened 2026-07-08
Track Referer and Origin headers on audit logs

dan2k3k4:
Persist the referer (query string and fragment stripped, so tokens/PII
from referring pages never reach the permanent log) and origin headers
on every audited request. Both surface in the admin audit log API and
UI, with a substring filter matching either column.

This PR adds Referer and Origin tracking to audit logs. The main changes are:

- New nullable `referer` and `origin` audit log columns.
- Middleware storage for sanitized header values.
- API response fields and substring filtering across both columns.
- Admin UI filter and separate Referer and Origin table columns.
- Tests for sanitization and referer/origin filtering.
</details>

This is close, but the header sanitizer should be fixed before merging.

- The migration chain and separate Origin column display look correct.
- The sanitizer still misses encoded userinfo in attacker-controlled headers.
- Those values are persisted and exposed through the admin audit log paths.

app/middleware/audit.py
</details>

<details open><summary><h3>Security Review</h3></summary>

Encoded credential-bearing Referer or Origin values can still be persisted to audit logs and shown through the admin API/UI.
</details>

| Filename | Overview |
|----------|----------|
| app/middleware/audit.py | Adds Referer and Origin sanitization before persistence, but encoded userinfo can still be stored. |
| app/api/audit.py | Adds Referer filtering across both header columns and includes both fields in responses. |
| app/migrations/versions/20260708_090000_a3f2b1c0d9e8_add_referer_origin_to_audit_logs.py | Adds nullable Referer and Origin columns on the current migration chain. |
| frontend/src/app/admin/audit-logs/_components/audit-log-table.tsx | Renders Referer and Origin as separate audit log table columns. |

</details>

<sub>Reviews (4): Last reviewed commit: ["docs: clarify referer/origin sanitizatio..."](https://github.com/amazeeio/amazee.ai/commit/784f25d7fb24fe0426c01463667d1fd71f49c843) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=42639667)</sub>

> Greptile also left **1 inline comment** on this PR.

### DIFF
--- a/app/api/audit.py
+++ b/app/api/audit.py
@@ -2,7 +2,7 @@ from fastapi import APIRouter, Depends, Query, HTTPException, status
 from sqlalchemy.orm import Session
 from typing import Optional
 from datetime import datetime
-from sqlalchemy import distinct, cast, String
+from sqlalchemy import distinct, cast, String, or_
 from app.db.database import get_db
 from app.api.auth import get_current_user_from_auth
 from app.schemas.models import (
@@ -31,11 +31,13 @@ async def get_audit_logs(
     from_date: Optional[datetime] = None,
     to_date: Optional[datetime] = None,
     status_code: Optional[str] = None,
+    referer: Optional[str] = None,
 ):
     """
     Retrieve audit logs with optional filtering.
     Only accessible by admin users.
     event_type, resource_type, and status_code can be comma-separated lists for multiple values.
+    referer matches as a substring against both the referer and origin columns.
     """
     if not current_user.is_admin:
         logger.warning(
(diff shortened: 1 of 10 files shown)


dspachos (approved):
Major issues: None blocking. 

Notes:
 - Sanitization is sound for the real threat model (tokens/PII in query strings/fragments/userinfo — exactly what the docstring promises). Greptile flagged "malformed values can persist credential-like text," but these headers are attacker-controlled, so there's no third-party data-exfil risk — the residual concern is an edge case. The "keep malformed raw for forensic value" tradeoff is reasonable.
 - Performance: the ilike '%…%' filter is a seq scan (can't use a btree index). It's acknowledged in a code comment ("add a pg_trgm index if audit table growth makes this slow") — fine for now, just be aware if the audit table gets large.

PR #619, opened 2026-07-14
chore: fix functional bugs found by deepsec scan

dan2k3k4:
Addresses all 11 BUG-severity findings from the deepsec scan. All functional (non-exploitable) bugs.

## Backend
- **Team-admin "Remove user" always 403'd** — `remove_user_from_team` was system-admin-only, but the button only appears for team admins. Relaxed to team-admin with a same-team ownership check (mirrors `update_user_role`); system admins unrestricted.
- **Admin PUT /users/{id} silently discarded password fields** — added a dedicated `AdminUserUpdate` schema (no password fields, `extra='forbid'`) so passwords now return a clear 422 instead of a silent no-op. Self-service password change still goes through /auth/me. (`is_active` added to the schema since callers already send it.)
- **ACCESS_TOKEN_EXPIRE_MINUTES ignored** — token lifetime now reads the setting instead of a hardcoded 60m.
- **/oauth2-redirect returned the full Swagger page** — now returns `get_swagger_ui_oauth2_redirect_html()`.
- **CREATE DATABASE unquoted identifier** — `wait_for_database.py` now quotes via `psycopg2.sql.Identifier` so hyphenated DB names produce valid DDL.

## CI
- **Release tag push failed under `contents: read`** — helm-package-publish job now has `contents: write`.

## Frontend
- **API tokens page** — consolidated on a single React Query source of truth: fixes the duplicate mount fetch, the never-shown delete spinner, and the newly created token not appearing in the list.
- **config client env** — stopped reading non-`NEXT_PUBLIC_` env vars (always undefined in the browser); fallbacks use inert defaults + the async /api/config load.
- **use-config** — awaits the in-flight load instead of returning an ambiguous null, so a concurrent caller no longer sees a spurious hard-error screen.
- **private-ai-keys-table** — copy before in-place sort so the `keys` prop is not mutated.

Full backend suite passes (1103 passed); changed frontend files typecheck clean.

This PR addresses 11 functional bugs identified by a security scan, spanning backend authorization, schema correctness, token expiry, frontend state management, and a CI permissions gap. All changes are targeted and well-scoped, with a full backend test suite passing and updated/new tests for the changed authorization logic.

- **Backend**: Fixes team-admin `remove_user_from_team` authorization (was always 403), swaps `UserUpdate` for a new `AdminUserUpdate` schema with `extra='forbid'` on the admin PUT endpoint (making silently-dropped password fields into explicit 422s, and activating `is_active` writes), corrects the `ACCESS_TOKEN_EXPIRE_MINUTES` being ignored, and returns the proper OAuth2 redirect HTML from `/oauth2-redirect`.
- **Frontend**: Consolidates API token page to React Query as the single source of truth (removing a duplicate manual fetch, a never-shown delete spinner, and missing post-create list refresh); stops reading non-`NEXT_PUBLIC_` env vars in the browser; deduplicates concurrent `loadConfig` calls via a shared in-flight promise; and copies the `keys` prop before sorting to avoid prop mutation.
- **Scripts / CI**: Quotes the database identifier in `wait_for_database.py` via `psycopg2.sql.Identifier` to handle hyphenated names, and upgrades the helm-publish job to `contents: write` so the release tag push succeeds.
</details>

The PR is safe to merge; the logic changes are well-targeted and backed by passing tests.

All backend fixes are correct and well-tested. The frontend token page consolidation is an improvement, but the shared deleteMutation.isPending across all token rows means every delete button enters loading state when any one token is deleted. The redundant double-fetch on mutation success and the newly functional is_active write path for team admins are worth a second look.

frontend/src/app/auth/token/page.tsx — shared delete loading state and double-fetch on mutation success; app/schemas/models.py — newly active is_active write path for team admins.
</details>

| Filename | Overview |
|----------|----------|
| frontend/src/app/auth/token/page.tsx | Consolidated to React Query as sole source of truth; removes manual fetch/state; redundant invalidate+refetch pattern and shared isPending across all delete buttons are P2 concerns. |
| app/schemas/models.py | Introduces AdminUserUpdate with extra='forbid'; correctly excludes password fields; newly active is_active field gives team admins ability to deactivate users — intentional but worth confirming. |
| app/api/users.py | Relaxes remove_user_from_team to team-admin with same-team ownership check; switches update_user to AdminUserUpdate; logic is correct and mirrors existing update_user_role pattern. |
| app/core/security.py | Fixes hardcoded 60-minute token expiry to read ACCESS_TOKEN_EXPIRE_MINUTES from settings; straightforward and correct. |
| app/main.py | Replaces incorrect get_swagger_ui_html call on /oauth2-redirect with get_swagger_ui_oauth2_redirect_html(); correct fix. |
| scripts/wait_for_database.py | Quotes database identifier via psycopg2.sql.Identifier, fixing DDL syntax for hyphenated names; clean and correct. |
| frontend/src/stores/use-config.ts | Introduces module-level inFlightLoad to deduplicate concurrent loadConfig calls; pattern is correct — inFlightLoad holds the Promise synchronously and is cleared in finally after settling. |
| frontend/src/utils/config.ts | Extracts fallbackConfig() helper and removes incorrect browser reads of non-NEXT_PUBLIC_ env vars; correct fix for always-undefined browser env reads. |
| frontend/src/components/private-ai-keys-table.tsx | Copies keys array before sorting to avoid mutating the prop; simple, correct fix. |
| .github/workflows/helm-package-publish.yml | Upgrades helm-package-publish job to contents: write so the release tag push no longer fails under read-only permissions. |
| tests/test_teams.py | Splits one incorrect test into two correct scenarios: team admin can remove same-team member (200) and cannot remove other-team member (403); coverage is thorough. |

</details>

<details><summary><h3>Flowchart</h3></summary>

<a href="#gh-light-mode-only">

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
    A["POST /users/id/remove-from-team"] --> B{get_role_min_team_admin}
    B -- "not authorized" --> Z[403 Forbidden]
    B -- "authorized" --> C{current_user.is_admin?}
    C -- "yes - system admin" --> E{db_user.team_id is None?}
    C -- "no - team admin" --> D{Same team?}
    D -- "no" --> Z
    D -- "yes" --> E
    E -- "yes - not in team" --> Y[400 Bad Request]
    E -- "no" --> F[Clear team_id in DB]
    F --> G[Sync LiteLLM]
    G --> I[200 User response]

    subgraph "PUT /users/id - AdminUserUpdate schema"
        K["is_admin=true by non-admin?"] -- "yes" --> Z2[403 Forbidden]
        K -- "no" --> L["Update: email, is_admin, is_active, receive_marketing_updates"]
    end
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
%%{init: {'theme': 'base', 'themeVariables': {"darkMode": true, "background": "#0d1117", "primaryColor": "#21262d", "primaryTextColor": "#e6edf3", "primaryBorderColor": "#8b949e", "lineColor": "#8b949e", "textColor": "#e6edf3", "edgeLabelBackground": "#161b22", "actorBkg": "#21262d", "actorBorder": "#8b949e", "actorTextColor": "#e6edf3", "actorLineColor": "#8b949e", "signalColor": "#8b949e", "signalTextColor": "#e6edf3", "noteBkgColor": "#373320", "noteBorderColor": "#d4a72c", "noteTextColor": "#f0e6c0", "labelBoxBkgColor": "#21262d", "labelBoxBorderColor": "#8b949e", "labelTextColor": "#e6edf3", "loopTextColor": "#e6edf3", "activationBkgColor": "#30363d", "activationBorderColor": "#8b949e"}}}%%
flowchart TD
    A["POST /users/id/remove-from-team"] --> B{get_role_min_team_admin}
    B -- "not authorized" --> Z[403 Forbidden]
    B -- "authorized" --> C{current_user.is_admin?}
    C -- "yes - system admin" --> E{db_user.team_id is None?}
    C -- "no - team admin" --> D{Same team?}
    D -- "no" --> Z
    D -- "yes" --> E
    E -- "yes - not in team" --> Y[400 Bad Request]
    E -- "no" --> F[Clear team_id in DB]
    F --> G[Sync LiteLLM]
    G --> I[200 User response]

    subgraph "PUT /users/id - AdminUserUpdate schema"
        K["is_admin=true by non-admin?"] -- "yes" --> Z2[403 Forbidden]
        K -- "no" --> L["Update: email, is_admin, is_active, receive_marketing_updates"]
    end
```

</a>
</details>

<details><summary><h3>Comments Outside Diff (3)</h3></summary>

1. `frontend/src/app/auth/token/page.tsx`, line 63-82 ([link](https://github.com/amazeeio/amazee.ai/blob/4a2e6385bcd1f8a994df4f09d9b89a087dfb2eea/frontend/src/app/auth/token/page.tsx#L63-L82)) 

   <a href="#">P2</a> **Shared loading state across all delete buttons**

   `deleteMutation` is a single mutation instance, so `deleteMutation.isPending` becomes `true` for all rendered `DeleteConfirmationDialog` components simultaneously when any one token deletion is in progress. If the dialog's `isLoading` prop disables the confirm button (which it appears to), every token's delete button is blocked while one is being deleted — and there's no per-token visual cue about which deletion is in flight. This is a UX regression from the previous per-token async handler (which had no loading state at all) and could confuse users with many tokens.

   A common fix is to track the in-flight token id in local state (e.g. `const [deletingId, setDeletingId] = useState<string | null>(null)`) and pass `isLoading={deleteMutation.isPending && deletingId === token.id}` per row.

   Note: If this suggestion doesn't match your team's coding style, reply to this and let me know. I'll remember it for next time!

2. `frontend/src/app/auth/token/page.tsx`, line 44-47 ([link](https://github.com/amazeeio/amazee.ai/blob/4a2e6385bcd1f8a994df4f09d9b89a087dfb2eea/frontend/src/app/auth/token/page.tsx#L44-L47)) 

   <a href="#">P2</a> Calling both `invalidateQueries` and `refetchQueries` is redundant. `invalidateQueries` already marks the `"tokens"` query stale and immediately triggers a refetch when there are active subscribers. The subsequent `refetchQueries` fires a second network request in parallel, doubling the API calls after every create or delete.

3. `frontend/src/app/auth/token/page.tsx`, line 67-70 ([link](https://github.com/amazeeio/amazee.ai/blob/4a2e6385bcd1f8a994df4f09d9b89a087dfb2eea/frontend/src/app/auth/token/page.tsx#L67-L70)) 

   <a href="#">P2</a> Same redundant double-fetch pattern on the delete mutation's `onSuccess`. `invalidateQueries` is sufficient.

</details>

<sub>Reviews (1): Last reviewed commit: ["chore: fix functional bugs found by deep..."](https://github.com/amazeeio/amazee.ai/commit/4a2e6385bcd1f8a994df4f09d9b89a087dfb2eea) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=44097641)</sub>

> Greptile also left **1 inline comment** on this PR.

### DIFF
--- a/.github/workflows/helm-package-publish.yml
+++ b/.github/workflows/helm-package-publish.yml
@@ -25,7 +25,7 @@ jobs:
   package-and-publish:
     runs-on: ubuntu-latest
     permissions:
-      contents: read
+      contents: write  # pushes the release tag in "Create Release Tag"
       packages: write
       security-events: write
 
--- a/app/api/auth.py
+++ b/app/api/auth.py
@@ -145,6 +145,11 @@ def create_and_set_access_token(
 
     # Set cookie expiration based on user role
     # System administrators get 8 hours (28800 seconds), regular users get 30 minutes (1800 seconds)
+    # NOTE: this cookie lifetime is deliberately independent of the JWT lifetime
+    # (settings.ACCESS_TOKEN_EXPIRE_MINUTES, applied in create_access_token). The
+    # effective session ends when the *shorter* of the two expires, so adjust
+    # both together — raising ACCESS_TOKEN_EXPIRE_MINUTES alone won't lengthen a
+    # regular user's session (this cookie dies first).
     cookie_expiration = 28800 if user and user.is_admin else 1800
 
     # Prepare cookie settings
--- a/app/api/users.py
+++ b/app/api/users.py
@@ -21,7 +21,7 @@ from app.db.database import get_db
 from app.core.dependencies import get_limit_service
 from app.schemas.models import (
     User,
-    UserUpdate,
+    AdminUserUpdate,
     UserCreate,
     TeamOperation,
     UserAdminRegionResponse,
@@ -899,7 +899,7 @@ async def get_user(
 )
 async def update_user(
     user_id: int,
-    user_update: UserUpdate,
+    user_update: AdminUserUpdate,
     current_user: DBUser = Depends(get_current_user_from_auth),
     db: Session = Depends(get_db),
 ):
@@ -1071,7 +1071,7 @@ async def add_user_to_team(
 @router.post(
     "/{user_id}/remove-from-team",
     response_model=User,
-    dependencies=[Depends(get_role_min_system_admin)],
+    dependencies=[Depends(get_role_min_team_admin)],
 )
 async def remove_user_from_team(
     user_id: int,
@@ -1079,11 +1079,20 @@ async def remove_user_from_team(
     db: Session = Depends(get_db),
 ):
     """
-    Remove a user from a team. Accessible by admin users.
+    Remove a user from a team. Accessible by system admins or team admins for
+    members of their own team.
     """
     db_user = db.query(DBUser).filter(DBUser.id == user_id).first()
     if not db_user:
         raise HTTPException(status_code=404, detail="User not found")
+
+    # Team admins may only remove members of their own team (mirrors
+    # update_user_role); system admins are unrestricted.
+    if not current_user.is_admin and db_user.team_id != current_user.team_id:
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="Not authorized to perform this action",
+        )
 
     # Check if user is a member of a team
     if db_user.team_id is None:
--- a/app/api/users.py
+++ b/app/api/users.py
@@ -1110,6 +1110,25 @@ async def remove_user_from_team(
             detail="User is not a member of any team",
         )
 
+    # Don't orphan a team: block removing its last remaining team admin (a team
+    # admin can now remove members, including themselves). Promote another admin
+    # first.
+    if db_user.role == UserRole.TEAM_ADMIN:
+        other_admins = (
+            db.query(DBUser)
+            .filter(
+                DBUser.team_id == db_user.team_id,
+                DBUser.id != db_user.id,
+                DBUser.role == UserRole.TEAM_ADMIN,
+            )
+            .count()
+        )
+        if other_admins == 0:
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail="Cannot remove the last team admin from the team",
+            )
+
     # Remove user from team
     previous_team_id = db_user.team_id
     db_user.team_id = None
--- a/app/api/users.py
+++ b/app/api/users.py
@@ -925,6 +925,15 @@ async def update_user(
                 detail="Team members cannot be made administrators",
             )
 
+    # Only system admins may change a user's activation status (mirrors the
+    # is_admin guard above). Team admins manage membership/role, not whether an
+    # account is active.
+    if user_update.is_active is not None and not current_user.is_admin:
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="Not authorized to perform this action",
+        )
+
     previous_email = db_user.email
     for key, value in user_update.model_dump(exclude_unset=True).items():
         setattr(db_user, key, value)
--- a/app/core/security.py
+++ b/app/core/security.py
@@ -45,7 +45,9 @@ def create_access_token(data: dict, expires_delta: Optional[timedelta] = None) -
     if expires_delta:
         expire = datetime.now(UTC) + expires_delta
     else:
-        expire = datetime.now(UTC) + timedelta(minutes=60)  # Default to 60 minutes
+        expire = datetime.now(UTC) + timedelta(
+            minutes=settings.ACCESS_TOKEN_EXPIRE_MINUTES
+        )
     to_encode.update({"exp": expire})
     encoded_jwt = jwt.encode(
         to_encode, settings.SECRET_KEY, algorithm=settings.ALGORITHM
--- a/app/main.py
+++ b/app/main.py
@@ -30,7 +30,10 @@ from app.middleware.prometheus import PrometheusMiddleware
 from fastapi import FastAPI
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.middleware.trustedhost import TrustedHostMiddleware
-from fastapi.openapi.docs import get_swagger_ui_html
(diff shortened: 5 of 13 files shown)


dspachos (approved):
### Things to confirm before merge

1. **`is_active` is a new team-admin capability, not just a schema tidy-up.** Previously `UserUpdate` had no `is_active`, so Pydantic silently discarded it (e.g. `tests/test_users.py:82` sends it but never actually tested persistence). Since PUT `/users/{id}` is `get_role_min_team_admin` with a same-team check, a team admin can now **deactivate/reactivate members of their own team — including reactivating a user a system admin deactivated, and deactivating themselves.** The PR body frames this as intentional, but it's a real authorization-surface expansion inside a bug-fix PR. If intended, it deserves a test (none of the new tests cover team-admin `is_active` writes); if not, gate it with the same `if not current_user.is_admin: 403` guard used for `is_admin`.

2. **Team admins can now remove themselves (or the other team admin) from their team**, since the same-team check passes for `db_user == current_user` and doesn't consider the target's role. This can orphan a team with no admin. `update_user_role` has the same shape, so it's consistent — but worth a deliberate decision (and possibly a "cannot remove the last team admin" guard as a follow-up).

### Minor / non-blocking

- **Cookie vs. token lifetime drift** (pre-existing, now more visible): `auth.py:148` hardcodes cookie `max_age` at 30 min / 8 h while the JWT now follows `ACCESS_TOKEN_EXPIRE_MINUTES`. Raising the setting won't extend sessions for regular users (cookie dies first), and admin cookies already outlive the 60-min JWT. Fine to leave, but a comment linking the two would prevent someone "fixing" one without the other.
- **Token page delete UX** (also flagged by Greptile, outside this diff's changed lines): the single `deleteMutation.isPending` puts every row's delete dialog into loading state at once, and `invalidateQueries` + `refetchQueries` in the mutation `onSuccess` handlers double-fetches. Both pre-date this PR — reasonable follow-up, not a blocker.
- **Test coverage gap**: no test asserts the new 422 when password fields hit PUT `/users/{id}` — that's the user-visible contract change from `extra='forbid'` and is cheap to add.

PR #649, opened 2026-07-29
feat(budget-alerts): notify MOAD when spend crosses budget thresholds (AI-448)

dspachos:
Customers get no warning before their AI budget runs out — keys simply stop working when
LiteLLM's `max_budget` is reached. This adds an amazee.ai-side job that detects when spend
crosses 50 / 75 / 90 / 100 % of budget and POSTs a batch of events to MOAD.

**MOAD's consumer endpoint and the customer-facing notification are a separate ticket.** This
PR only produces and delivers the events. It ships disabled (`BUDGET_ALERT_ENABLED=false`).

## Why amazee.ai computes the percentage instead of LiteLLM

LiteLLM has native budget alerting, but it cannot produce the right number for our teams:

- **Wrong numerator.** LiteLLM's team-level `spend` counter is never reset; our billing cycle
  zeroes the *per-key* spends instead. A lifetime total over a one-cycle budget sits
  permanently above 100 %.
- **Moving denominator.** POOL budgets are owned by our ledger. Buying a top-up raises the
  budget, so a percentage can fall back below a band it already crossed. LiteLLM's TTL-based
  dedup would then suppress the genuine second crossing.
- Its thresholds are hardcoded to 85 % / 95 %, and there is one webhook URL per proxy.

So LiteLLM is the source of *spend*, and every budget number comes from our DB.

## How a tick works

`monitor-budget-thresholds` runs every 5 minutes (`.lagoon.yml`, prod + dev) via
`scripts/trigger_budget_alerts_job.py`, guarded by the `budget_alerts` advisory lock.

1. **Pick regions** — `regions_to_sweep()`: regions with at least one live key and LiteLLM
   credentials. Deliberately *not* filtered on `is_active`; PROD region 5 is inactive but
   holds 11,859 of 13,930 keys.
2. **One wide read per region** — unfiltered `/team/daily/activity` over the window that
   `region_sweep_start()` derives from the ledger. LiteLLM writes no daily rows for idle
   entities, so an 11,859-key region reports 42 active teams and 351 active keys.
3. **One scoped read per active team** — `/key/list?team_id=` for `max_budget`, cap duration
   and reset date. Its `spend` figures are deliberately ignored.
4. **Compute, compare, fire** — only the highest newly-crossed band per subject.
5. **POST the batch**, then advance state only for the events the POST accepted.
6. **Audit + metrics** — a `DBAuditLog` row per delivered crossing, 5 Prometheus series.

Subjects do not share a window, so the daily rows are indexed once per tick as a per-day series
and each subject sums only the days its own cycle covers.

Cost is `pages + active_teams` calls per region, independent of idle key count. Measured ~1 s
per region. Sweeping every key instead — the shape of the existing hourly reconciler, ~40 min
at 13.9k keys — is what does not survive 100k.

## Scope

POOL teams only, and keys carrying **both** a team and a region — the shape MOAD provisions
and can address. Both service keys and user keys. PERIODIC is excluded because all 12,098
PERIODIC keys on PROD belong to the single anonymous Drupal trial team, which has no MOAD
workspace and no addressable owner.

Subjects: **team per region**, **key**, **team member**.

## Windows: alerts are per cycle

Every percentage is spend over the **current cycle** divided by that cycle's budget. Which
cycle depends on the subject, because a team and a capped key are not on the same clock.

| Subject | Window |
| --- | --- |
| Team with a subscription | the subscription cycle (`effective_period_start` → now) |
| Team with top-ups only | the **oldest still-valid** purchase → now (window closes 365d after the *newest* purchase) |
| Team with **both** | the **subscription** cycle, against the whole available balance |
| Key or member **with a cap** | that cap's own current cycle |
| Key bounded by the pool | the team's window — it has no cycle of its own |

PROD: 88 subscription-backed POOL team+regions, 222 top-up-only.

**Subscriptions — never wider than the cycle**, because the denominator is
`purchased − consumed_cents` and `consumed_cents` is incremented at cycle close while LiteLLM's
spend is reset on the same call (invariant at `app/api/spend.py:982-987`). A closed cycle's spend
is already out of the budget; counting it again inflates the percentage and fires alerts nobody
earned.

**Top-ups — the opposite**, because `consumed_cents` is *never* written for them: FIFO allocation
runs at invoice close and a team with no subscription never has one. On PROD exactly **1 of 233**
top-up entries has it set, so the budget is the full face value of every still-valid purchase. A
window opening at the newest purchase would leave spend against the older entries counted
nowhere — missing from the numerator and never deducted from the budget.

Worked example: $30 bought 6 July with $25 spent against it, then $30 more on 29 July. Anchored on
the newest purchase that reads **$0 of $60** while the team really holds $35, and later reports
50 % when the truth is 92 % — told they are halfway through while their keys are about to stop.
Anchored on the oldest it reads $25 of $60 and keeps counting. Exposure: 7 of 222 top-up
team+regions hold more than one valid entry, only 3 bought on different days.

**Why "still-valid" carries the weight.** Settlement re-bases rather than annotates:
`materialize_topup_rollovers` deactivates a partly-consumed entry and writes a fresh
`topup_rollover` holding the remainder with `consumed_cents = 0`. So among valid entries the
column is 0, face value equals remaining, and the window lines up with the budget by
construction — already-settled spend is counted on neither side.

Cancellation is the exception: `app/api/subscription.py` runs `allocate_period_spend_fifo` but
never calls `materialize_topup_rollovers`, leaving `consumed_cents` on entries that stay valid.
The budget would then drop by spend the numerator still counts — subtracted once and added once —
and read high. `_unsettled_consumed_dollars` adds it back, a no-op unless the column is set, and
logs when it is not. A cancelled subscription otherwise falls through to the top-up window
cleanly, since deactivation sets `is_active = False` on every subscription row.

Cycle starts come from the shared `resolve_team_period_window()` — the same helper the spend API
uses — so an alert cannot quote a different period than the dashboard. `team_spend_periods` was
rejected as the source: 145 rows, only 4 covering the present.

Caps are per-cycle allowances, not absolute budgets: all 196 key caps on PROD are `31d` (130) or
`1mo` (66), and all 38 member caps are `1mo`. A `31d` cycle takes its phase from LiteLLM's
`budget_reset_at` when the durations agree — its reset job recomputes each boundary from the day
it runs, so that date absorbs any lateness — else from the cap row's own `created_at`, rolled
forward to the cycle containing now. `1mo` follows the calendar month, because LiteLLM really
does reset those keys on the 1st and the alert has to predict the enforcement the key is under.

## Other cases worth knowing

| Case | Resolution |
| --- | --- |
| LiteLLM key `spend` counters | Never a numerator — a top-up does not reset them, so they are lifetime totals |
| Subscription cancelled | Subscription rows are deactivated, so the top-up window takes over. Any `consumed_cents` FIFO left on still-valid entries is added back, since the numerator still counts that spend |
| Top-up bought mid-cycle | `period_start` does not move, so `period_key` is stable. The budget rises, the percentage falls, the band walks back silently and `arm_seq` is bumped so a real re-crossing still notifies. Bands are not blanket-reset, so a team still above 50 % is not warned twice |
| Expiring key (`0d`) | Spend still counts towards team and owner; only the key-scope alert is suppressed |
| Team window opens before the sweep floor | Back-filled with a scoped per-team activity call; skipped only if that call fails |
| POOL team with no valid ledger entry | No team budget. A `limited_resources` BUDGET row is a provisioning allowance, not credit (673 such teams) |
| Uncapped key, team has >1 key | Measured against the pool — answers "which key is eating the pool" |
| Uncapped key, team has 1 key | Withheld; arithmetically identical to the team event |
| `max_budget` 0 or null | Skipped — pool-gate-blocked keys would otherwise emit 100 % at creation |
| Idle team whose budget just fell | Found by `denominator_change_candidates`, not by traffic |
| 0 % → 95 % in one tick | One event (90), not three |
| Delivery fails | State unadvanced → retried next tick under the **same** `event_id` |
| Genuine re-crossing after a re-arm | **New** `event_id` via `arm_seq`, so dedup cannot swallow it |

`event_id` is `sha256(subject_key|period_key|band|arm_seq)`, not a uuid: a retry must carry the
id of the attempt it repeats, while a real second crossing of the same band must not.

## Payload

```json
{
  "api_version": "2026-08-01",
  "events": [{
    "event_id": "evt_9f2c…",
    "type": "budget.threshold_reached",
    "created_at": "2026-07-30T10:04:11+00:00",
    "data": {
      "scope": "team",
      "threshold_percent": 90,
      "percent_used": 92.14,
      "spend": 92.14,
      "max_budget": 100.0,
      "budget_source": "team_ledger",
      "currency": "USD",
      "budget_duration": "31d",
      "region": {"id": 6, "name": "amazeeai-us103"},
      "team": {"id": 719, "name": "Acme"},
      "user": null,
      "key": null,
      "period": {"key": "…", "start": "…", "end": "…"}
    }
  }]
}
```

`budget_source` tells the consumer how to word the message: `key_cap` / `litellm_key_budget`
(the key's own limit), `team_pool` (uncapped key measured against the pool), `team_ledger`
(team or member scope). The LiteLLM token is never sent — it is a live credential and `key.id`
identifies the key.

## Anti-drift

`resolve_team_period_window()` moves to `app/core/spend_period_service.py` and `get_team_spend`
now calls it, so the spend API and the alert engine cannot disagree about a period. That step
is behaviour-neutral; existing spend tests pass unchanged.

## Config

| Setting | Default |
| --- | --- |
| `BUDGET_ALERT_ENABLED` | `false` |
| `BUDGET_ALERT_THRESHOLDS` | `50,75,90,100` |
| `BUDGET_ALERT_WEBHOOK_URL` | `""` — full URL, so the route stays MOAD's choice |
| `BUDGET_ALERT_WEBHOOK_TOKEN` | falls back to `MOAD_DASHBOARD_API_TOKEN` |
| `BUDGET_ALERT_BATCH_SIZE` | `200` |
| `BUDGET_ALERT_REGION_CONCURRENCY` | `4` |
| `BUDGET_ALERT_MAX_LOOKBACK_DAYS` | `370` (floor on the one wide request; older windows are back-filled per team) |
| `BUDGET_ALERT_RECHECK_GRACE_HOURS` | `24` |

## Migration

`d1e2f3a4b5c6` adds `budget_alert_state` — **one row per subject**, overwritten in place, so
the table is bounded by subject count rather than growing with events. DEV is still at
`6c201dbaea6e` and the table does not exist yet, so this migration has never run.

## Testing

- 70 tests in `tests/test_budget_alerts.py` plus 3 window tests in
  `tests/test_spend_period_service.py`; full suite 1237 passed / 2 skipped; ruff clean.
- Note: the suite passed *unchanged* before those 3 window tests were added — nothing had been
  pinning the spend API's top-up window, so a customer-visible rule was uncovered by 1231
  tests.
- Comments and docstrings carry no environment specifics (no region names, team ids or key
  counts). The measured figures behind the decisions live in this description and in the
  design notes instead, where they can go stale without misleading anyone reading the code.
- Both numerator paths validated to the cent against the live DEV spend API.
- Sweep cost and the active-entity ratios measured on PROD region 5 (read-only).

## Not done yet

**This has never run end to end.** Every number above is validated against live data, but the
job itself has not executed. Roll out with `BUDGET_ALERT_ENABLED=false`, confirm the logged
percentages against `GET /spend/{region_id}/team/{team_id}`, then set the URL and enable.

Two things worth knowing rather than fixing here:

- **Member alerts will be rare.** PROD has 38 `team_member` spend caps and 54 POOL users with
  a manual USER BUDGET limit, against 1,345 POOL users. No cap means no line to cross.
- **`team`-scope spend caps have no PROD data** (0 rows), so that tightening path is only
  covered by unit tests.

Out of scope: MOAD's consumer endpoint and UI, non-budget COUNT/RPM limits, per-team
thresholds, HMAC signing, a retry outbox, and forecast ("projected to exceed") alerts.

## Follow-ups found while reviewing, deliberately not in this PR

1. **`app/api/spend.py:1334-1357` never rolls its cap anchor**, so for a long-lived team it
   reports a capped key's `period_start` far in the past and a `budget_reset_at` that has
   already passed. The amount is right, the dates are not. Fixing it changes a customer-visible
   number — `current_cycle_start()` is now available to reuse.
2. **104 caps still carry `1mo`/`30d`** (66 key, 38 member) so they reset on the 1st rather than
   from the day they were set. Every `1mo` row stops at 2026-06-15 and every `31d` row starts
   2026-06-17, so it is a clean historical cutover. Migrating them needs no engine change.
3. **Team-cap duration is stored inconsistently**: for a purchase-gated POOL team,
   `app/api/spend.py:1796` stores `"1mo"` while pushing `31d`/`<days-left>d` to LiteLLM. No PROD
   impact today (0 team-scope caps), but the engine reads the stored value.
4. **A top-up does not extend existing credit.** `add_topup_entry` sets the expiry of the new
   entry only, so older credit lapses first — 3 PROD teams hold entries with different expiry
   dates. Reviewed and deliberately left as-is; changing it is a product decision about how long
   credit lives, not an alerting one.

## Behaviour change to the spend API

`GET /spend/{region_id}/team/{team_id}` reports `period_start` earlier, and a higher
`period_spend`, for top-up teams holding more than one still-valid purchase — 3 teams on PROD
(13536, 4863, 7627). That is the correct figure: today those teams are shown a spend number that
omits money they actually spent. `period_end` is unchanged (newest purchase + 365d), and the 215
single-purchase teams see no difference at all.

This PR adds periodic budget-threshold monitoring and MOAD event delivery.
- Computes team, key, and member spend against subject-specific budget cycles.
- Uses deterministic event identifiers and persisted arming state for retry-safe delivery and genuine re-crossings.
- Backfills scoped activity for windows older than the regional sweep and discovers historical-only teams from ledger state.
- Adds configuration, scheduling, metrics, audit records, webhook delivery, database state, and coverage for budget-alert behavior.
</details>

The PR appears safe to merge.

No blocking failure remains; the previously reported retry-identity, re-arming, lookback truncation, backfill suppression, and historical-only discovery issues are addressed in the current implementation.
</details>

| Filename | Overview |
|----------|----------|
| app/core/budget_alert_service.py | Implements regional discovery, cycle-aware spend evaluation, deterministic crossing identity, scoped historical backfill, delivery-state advancement, metrics, and auditing. |
| app/core/spend_period_service.py | Centralizes team period-window and cap-cycle calculations shared by the spend API and alert engine. |
| app/services/budget_alert_webhook.py | Serializes threshold events and delivers bounded batches with all-or-nothing acceptance per HTTP request. |
| app/services/litellm.py | Adds paginated regional activity and team-scoped key/activity reads used by the alert sweep. |
| app/db/models.py | Adds bounded per-subject alert state, including the arming sequence required to distinguish retries from genuine re-crossings. |
| app/migrations/versions/20260729_220000_d1e2f3a4b5c6_add_budget_alert_state.py | Creates the budget alert state table and indexes needed by subject classification. |
| scripts/trigger_budget_alerts_job.py | Adds the lock-guarded scheduled-job entry point for threshold monitoring. |
| tests/test_budget_alerts.py | Covers threshold classification, retry identity, re-arming, historical backfill, delivery handling, and subject-specific budget windows. |

</details>

<details open><summary><h3>Sequence Diagram</h3></summary>

```mermaid
sequenceDiagram
    participant Cron
    participant AlertJob
    participant DB
    participant LiteLLM
    participant MOAD

    Cron->>AlertJob: Run every five minutes
    AlertJob->>DB: Acquire budget_alerts lock
    AlertJob->>DB: Resolve regions, budgets, windows, and prior state
    AlertJob->>LiteLLM: Fetch regional daily activity
    opt Window predates regional sweep
        AlertJob->>LiteLLM: Fetch scoped team activity
    end
    AlertJob->>LiteLLM: Fetch active-team key limits
    AlertJob->>AlertJob: Compute highest newly crossed bands
    AlertJob->>DB: Persist silent resets and increment arm sequence
    AlertJob->>MOAD: POST event batches
    MOAD-->>AlertJob: Accept delivered batches
    AlertJob->>DB: Advance only accepted event state and write audit rows
    AlertJob->>DB: Release lock
```
</details>

<sub>Reviews (9): Last reviewed commit: ["fix(budget-alerts): discover teams whose..."](https://github.com/amazeeio/amazee.ai/commit/10fb63231227b9c78309bfe9fd6bc987f4095e17) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=48594837)</sub>

<details><summary><h4>Context used (3)</h4></summary>

- Knowledge Base — [Budgets and Spend Limits](https://app.greptile.com/amazee-io/-/custom-context/knowledge-base/amazeeio/amazee.ai/-/docs/budgets-spend.md)
- Knowledge Base — [Database models, session, and migrations](https://app.greptile.com/amazee-io/-/custom-context/knowledge-base/amazeeio/amazee.ai/-/docs/database-models.md)
</details>

### DIFF
--- a/.lagoon.yml
+++ b/.lagoon.yml
@@ -27,6 +27,13 @@ environments:
         schedule: "30 2 * * *"
         command: python scripts/trigger_prune_signup_events_job.py
         service: backend
+      # Frequent on purpose: a 90% warning is worthless if it lands an hour after
+      # the key stopped working. The sweep is 2 LiteLLM calls per region and
+      # scales with active entities, not key count, so 5 minutes is affordable.
+      - name: monitor-budget-thresholds
+        schedule: "*/5 * * * *"
+        command: python scripts/trigger_budget_alerts_job.py
+        service: backend
     routes:
       - backend:
         - api.amazee.ai:
@@ -65,6 +72,10 @@ environments:
       - name: prune-signup-events
         schedule: "30 2 * * *"
         command: python scripts/trigger_prune_signup_events_job.py
+        service: backend
+      - name: monitor-budget-thresholds
+        schedule: "*/5 * * * *"
+        command: python scripts/trigger_budget_alerts_job.py
         service: backend
 
 tasks:
--- a/app/api/spend.py
+++ b/app/api/spend.py
@@ -1,5 +1,4 @@
 import logging
-import re
 from datetime import UTC, date, datetime, timedelta
 
 from fastapi import APIRouter, Depends, HTTPException, status
@@ -22,6 +21,10 @@ from app.core.security import (
     get_current_user_from_auth,
     get_private_ai_access,
     get_role_min_team_admin,
+)
+from app.core.spend_period_service import (
+    compute_period_start,
+    resolve_team_period_window,
 )
 from app.db.database import get_db
 from app.db.models import (
@@ -65,51 +68,10 @@ logger = logging.getLogger(__name__)
 MONTHLY_BUDGET_DURATION = "1mo"
 
 
-def _compute_period_start(
-    budget_reset_at: datetime | None, budget_duration: str | None
-) -> datetime | None:
-    """
-    Derive the start of the current budget period from LiteLLM's
-    ``budget_reset_at`` (end-of-period) and ``budget_duration``.
-
-    LiteLLM sets ``budget_reset_at`` to the moment the budget will auto-reset.
-    For ``"Nd"`` durations the reset is rolling N days after the last update;
-    for ``"1mo"`` / ``"30d"`` it snaps to the 1st of the next calendar month.
-
-    We parse the duration string and subtract from ``budget_reset_at`` to get
-    a best-effort calendar ``period_start``.  Returns ``None`` when either
-    input is missing or the duration cannot be parsed.
-    """
-    if budget_reset_at is None or not budget_duration:
-        return None
-
-    # Handle "1mo" / "30d" — both snap to 1st of next calendar month
-    # so the period start is always the 1st of the current month.
-    if budget_duration in ("1mo", "30d"):
-        # budget_reset_at is midnight on the 1st of next month.
-        # If reset is on 1st, the period that just ended started last month.
-        if budget_reset_at.day == 1:
-            if budget_reset_at.month == 1:
-                return budget_reset_at.replace(
-                    year=budget_reset_at.year - 1, month=12, day=1
-                )
-            return budget_reset_at.replace(month=budget_reset_at.month - 1, day=1)
-        return budget_reset_at.replace(day=1)
-
-    match = re.fullmatch(r"(\d+)([dhms])", budget_duration)
-    if not match:
-        return None
-    value = int(match.group(1))
-    unit = match.group(2)
-    if unit == "d":
-        return budget_reset_at - timedelta(days=value)
-    if unit == "h":
-        return budget_reset_at - timedelta(hours=value)
-    if unit == "m":
-        return budget_reset_at - timedelta(minutes=value)
-    if unit == "s":
-        return budget_reset_at - timedelta(seconds=value)
-    return None
+# Period-window maths lives in spend_period_service so the budget-threshold
+# alert engine derives the same window as this API. Kept under the original
+# private name because callers (and tests) already import it from here.
+_compute_period_start = compute_period_start
 
 
 @router.get(
@@ -973,64 +935,15 @@ async def get_team_spend(
     # For subscription-managed POOL teams, expose cycle window semantics from
     # Amazee DB ledger truth instead of LiteLLM counters.
     if team.budget_type == BudgetType.POOL:
-        now = datetime.now(UTC)
-        active_subscription = (
-            db.query(DBPeriodicBudgetLedgerEntry)
-            .filter(
-                DBPeriodicBudgetLedgerEntry.team_id == team_id,
-                DBPeriodicBudgetLedgerEntry.region_id == region_id,
-                DBPeriodicBudgetLedgerEntry.entry_type == "subscription",
-                DBPeriodicBudgetLedgerEntry.is_active.is_(True),
-                DBPeriodicBudgetLedgerEntry.effective_period_start.isnot(None),
-                DBPeriodicBudgetLedgerEntry.effective_period_end.isnot(None),
-                DBPeriodicBudgetLedgerEntry.effective_period_end > now,
-            )
-            .order_by(
-                DBPeriodicBudgetLedgerEntry.effective_period_end.desc(),
-                DBPeriodicBudgetLedgerEntry.id.desc(),
-            )
-            .first()
-        )
+        pool_window = resolve_team_period_window(db, team, region_id)
+        team_budget_duration = pool_window.budget_duration
+        team_budget_reset_at = pool_window.period_end
+        team_period_start = pool_window.period_start
+        active_subscription = pool_window.active_subscription
         if active_subscription is not None:
-            team_budget_duration = "31d"
-            team_budget_reset_at = active_subscription.effective_period_end
-            team_period_start = active_subscription.effective_period_start
             # For POOL teams, expose only current-cycle subscription amount.
             sub_cycle_budget_cents = int(active_subscription.amount_cents or 0)
             periodic_budget_view = round(sub_cycle_budget_cents / 100.0, 4)
-        else:
-            active_topup = (
-                db.query(DBPeriodicBudgetLedgerEntry)
-                .filter(
-                    DBPeriodicBudgetLedgerEntry.team_id == team_id,
-                    DBPeriodicBudgetLedgerEntry.region_id == region_id,
-                    DBPeriodicBudgetLedgerEntry.entry_type.in_(
-                        ["topup", "topup_rollover"]
-                    ),
-                    DBPeriodicBudgetLedgerEntry.is_active.is_(True),
-                    (
-                        DBPeriodicBudgetLedgerEntry.expires_at.is_(None)
-                        | (DBPeriodicBudgetLedgerEntry.expires_at > now)
-                    ),
-                )
-                .order_by(
-                    DBPeriodicBudgetLedgerEntry.purchased_at.desc(),
-                    DBPeriodicBudgetLedgerEntry.id.desc(),
-                )
-                .first()
-            )
-            pool_duration = f"{settings.POOL_PURCHASE_EXPIRY_DAYS}d"
-            team_budget_duration = pool_duration
-            if active_topup is not None and active_topup.purchased_at is not None:
-                anchor = active_topup.purchased_at
-            else:
-                anchor = team.created_at or now
-            if anchor.tzinfo is None:
-                anchor = anchor.replace(tzinfo=UTC)
-            team_period_start = anchor
-            team_budget_reset_at = anchor + timedelta(
-                days=settings.POOL_PURCHASE_EXPIRY_DAYS
-            )
 
     if team.budget_type != BudgetType.POOL:
         periodic_budget_view = None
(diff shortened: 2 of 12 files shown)


pmelab (approved):
Approved! There might be still scaling issues if teams get a large number of keys, but we can cross that bridge when we get there.

PR #657, opened 2026-08-03
feat(trial): reap abandoned trial keys and guard bulk deletion

dspachos:
Ticket: amazeeio/moad#668

## Why

Every anonymous trial provisions a LiteLLM key **and** a Postgres vector database. Nothing ever reclaimed either — `monitor_trial_users` runs monthly and only expires keys of over-budget users, leaving the key row, the user row and the database behind. de103 reached **11,507** trial keys that way. de-eu101 was filling at the same rate: 1,129 keys and 1,129 databases in its first 7 days as the trial region.

On 2026-08-02 an ad-hoc cleanup script made the other half of the problem visible. It walked every region and expired **1,110 keys that had just been minted on the new trial region**, some only days old. No committed script existed to do this safely, and the nearest template — `scripts/restrict_trial_key_routes.py` — has the same all-regions shape, so the next script written from it would repeat the fault.

### `app/core/trial_cleanup.py` (new)

Holds both safety rules so the manual script and the scheduled reaper cannot drift apart.

**Region guard.** Selection is scoped to one region and refuses an *unfiltered* sweep of whatever `AI_TRIAL_REGION` resolves to. The live region stays reapable — it is the one that actually accumulates, so excluding it would make the reaper pointless — but only for keys older than `LIVE_REGION_MIN_AGE_DAYS` (7) with no recorded spend. A key issued to someone who signed up this week can never be taken. The guard lives inside `select_trial_keys`, so there is no way to obtain a list of keys to delete without it running.

**Deletion order.** A key row is deleted only once its LiteLLM key and vector database are gone. If either remote call fails, every row stays in place for a later retry. This is deliberately the opposite of `hard_delete_expired_teams`, which logs remote failures and drops the rows regardless — that path strands the resources with nothing left pointing at them. Both remote calls are idempotent (`404 → True`, `DROP ... IF EXISTS`), so retrying a partial delete is safe.

Also clears `spend_caps.key_id` **and** `budget_alert_state.key_id` — both are RESTRICT FKs to `ai_tokens.id`. `audit_logs.user_id` is detached rather than deleted, so the signup record keeps its details, IP and timestamp.

### `reap_trial_keys` + `reap-trial-keys` cron

Daily at 01:00. Deletes trial keys older than `AI_TRIAL_RETENTION_DAYS` (default 30) with no recorded spend, along with their LiteLLM key, vector database, key row and user row. Capped per run by `AI_TRIAL_REAP_BATCH_SIZE` (default 500) so a first pass over a large backlog is spread over several nights. Stops on a region after 10 failures — that means the host is down, not that the keys are bad.

Setting `AI_TRIAL_RETENTION_DAYS` below the live region's 7-day minimum makes the reaper skip the live region with a warning rather than mass-delete.

### `scripts/cleanup_trial_keys.py` (new)

Manual, region-targeted clean-up over the same core. Requires `--region <id>`, dry-run by default, `--apply` to act, confirmation prompt unless `--yes`.

### Two silent failures in the signup path

- The failure handler deleted the user row and committed *before* attempting the LiteLLM delete. When that second step failed it left a working key with no row pointing at it — invisible to every cleanup path, since all of them iterate rows. One such key was found on de103, minted 2026-04-07 and still valid until 2027. The LiteLLM delete now goes first.
- An unresolvable `AI_TRIAL_REGION` 404s every trial signup and logged nothing. A rename or deactivation was only visible as signup volume dropping to zero. It now logs at error level.

## Testing

`make backend-test` — **1287 passed, 2 skipped**. Ruff clean.

32 new tests across `tests/test_trial_cleanup.py` and `tests/test_reap_trial_keys.py`, including the 2026-08-02 incident as a regression test, the live-region boundary in both directions, and dead-endpoint cases asserting that no row is removed.

## Notes for review

- **Retention default of 30 days is a product call.** 99.6% of trial keys are never used at all (46 of 11,269 on de103 ever spent anything, $17.02 total), so 30 days clears essentially everything while leaving a real trial window intact.
- **The reaper deletes trial user rows**, not just keys. Team 719 holds 12,638 users and only grows. Say if you would rather it left them.
- **Once deployed with a 30-day retention, the reaper will delete the 1,110 already-expired region-67 keys** around 2026-08-27, with their vector databases. If those should be reissued instead, that needs deciding first.
- **Out of scope, still open:** `private_ai_keys.py:1040` (the API delete endpoint) clears `spend_caps` but not `budget_alert_state`. That table is not in prod yet so the de103 drain is safe today, but the drain runs through that endpoint and the two will eventually overlap.

The PR adds scheduled and manual cleanup for abandoned trial keys while preserving retryable database pointers until remote resources are removed.
- Adds region-scoped trial-key selection with age and operator safeguards.
- Deletes LiteLLM keys and vector databases before local key and optional user rows.
- Adds a locked daily reaper, a dry-run-first cleanup CLI, signup cleanup ordering, and regression coverage.
</details>

The PR appears safe to merge.

No blocking failure remains; the SQL-level unused-key filtering resolves the batch-stall thread, and the manual former-region path now assesses recent or spent keys and requires an explicit destructive override.
</details>

| Filename | Overview |
|----------|----------|
| app/core/trial_cleanup.py | Introduces shared region selection, risk assessment, remote-first deletion, and local cascade cleanup; the previously reported selection-limit and former-region CLI safeguards are addressed. |
| app/core/worker.py | Adds the scheduled, age-based trial-key reaper with per-region service reuse, bounded batches, and failure cutoffs. |
| scripts/cleanup_trial_keys.py | Adds a region-required, dry-run-first manual cleanup command with explicit gates for recent or spent keys. |
| scripts/trigger_trial_reaper_job.py | Adds the locked cron entry point for the new reaper. |
| app/api/auth.py | Improves unresolved-region diagnostics and deletes a newly created LiteLLM key before removing its local user pointer during failed signup cleanup. |
| tests/test_trial_cleanup.py | Covers selection guards, spend filtering, deletion ordering, retry behavior, and dependent-row cleanup. |
| tests/test_reap_trial_keys.py | Covers retention boundaries, spent-key policy, remote failures, batching, and user deletion for the scheduled reaper. |

</details>

<details open><summary><h3>Sequence Diagram</h3></summary>

```mermaid
sequenceDiagram
    participant Scheduler
    participant Reaper
    participant DB
    participant LiteLLM
    participant Postgres
    Scheduler->>Reaper: Run locked daily cleanup
    Reaper->>DB: Select region-scoped keys older than retention
    loop Each selected key
        Reaper->>LiteLLM: Delete key
        LiteLLM-->>Reaper: Success or idempotent not-found
        Reaper->>Postgres: DROP database/role if present
        Postgres-->>Reaper: Success
        Reaper->>DB: Delete dependent rows, key, and eligible user
        Reaper->>DB: Commit
    end
```
</details>

<sub>Reviews (4): Last reviewed commit: ["docs(trial): keep code comments general,..."](https://github.com/amazeeio/amazee.ai/commit/01b162cccf2ab24d5a301be01a9ffda86888bf5e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=49608056)</sub>

**Context used:**

- Knowledge Base — [Background/Periodic Jobs (worker + internal trigger API)](https://app.greptile.com/amazee-io/-/custom-context/knowledge-base/amazeeio/amazee.ai/-/docs/worker-background-jobs.md)

### DIFF
--- a/.lagoon.yml
+++ b/.lagoon.yml
@@ -19,6 +19,13 @@ environments:
         schedule: "0 5 1 * *"
         command: python scripts/trigger_trial_recon_job.py
         service: backend
+      # Daily, not monthly: every trial provisions a LiteLLM key AND a Postgres
+      # database, and nothing else reclaims either. Monthly lets ~6,000 of each
+      # pile up between runs. Capped per run by AI_TRIAL_REAP_BATCH_SIZE.
+      - name: reap-trial-keys
+        schedule: "0 1 * * *"
+        command: python scripts/trigger_trial_reaper_job.py
+        service: backend
       - name: refresh-disposable-domains
         schedule: "0 2 * * *"
         command: python scripts/trigger_refresh_disposable_domains_job.py
@@ -64,6 +71,13 @@ environments:
       - name: monitor-trial-users
         schedule: "0 5 1 * *"
         command: python scripts/trigger_trial_recon_job.py
+        service: backend
+      # Daily, not monthly: every trial provisions a LiteLLM key AND a Postgres
+      # database, and nothing else reclaims either. Monthly lets ~6,000 of each
+      # pile up between runs. Capped per run by AI_TRIAL_REAP_BATCH_SIZE.
+      - name: reap-trial-keys
+        schedule: "0 1 * * *"
+        command: python scripts/trigger_trial_reaper_job.py
         service: backend
       - name: refresh-disposable-domains
         schedule: "0 2 * * *"
--- a/.lagoon.yml
+++ b/.lagoon.yml
@@ -19,9 +19,6 @@ environments:
         schedule: "0 5 1 * *"
         command: python scripts/trigger_trial_recon_job.py
         service: backend
-      # Daily, not monthly: every trial provisions a LiteLLM key AND a Postgres
-      # database, and nothing else reclaims either. Monthly lets ~6,000 of each
-      # pile up between runs. Capped per run by AI_TRIAL_REAP_BATCH_SIZE.
       - name: reap-trial-keys
         schedule: "0 1 * * *"
         command: python scripts/trigger_trial_reaper_job.py
@@ -72,9 +69,6 @@ environments:
         schedule: "0 5 1 * *"
         command: python scripts/trigger_trial_recon_job.py
         service: backend
-      # Daily, not monthly: every trial provisions a LiteLLM key AND a Postgres
-      # database, and nothing else reclaims either. Monthly lets ~6,000 of each
-      # pile up between runs. Capped per run by AI_TRIAL_REAP_BATCH_SIZE.
       - name: reap-trial-keys
         schedule: "0 1 * * *"
         command: python scripts/trigger_trial_reaper_job.py
--- a/app/core/config.py
+++ b/app/core/config.py
@@ -64,9 +64,9 @@ class Settings(BaseSettings):
     # LiteLLM key AND a Postgres database, and nothing else ever reclaims
     # either, so without this they accumulate for the life of the region.
     #
-    # 90 days, not 30: someone can set up a Drupal site and not touch its AI
-    # key for a month or two, so a shorter window would delete keys people are
-    # still going to use.
+    # Generous on purpose: someone can set up a site and not touch its AI key
+    # for a month or two, so a short window would delete keys people are still
+    # going to use.
     AI_TRIAL_RETENTION_DAYS: int = int(os.getenv("AI_TRIAL_RETENTION_DAYS", "90"))
     # Cap per reaper run. Each deletion is an HTTP call plus a DROP DATABASE,
     # so a first run against a large backlog is spread over several nights
--- a/app/core/config.py
+++ b/app/core/config.py
@@ -60,6 +60,14 @@ class Settings(BaseSettings):
         "AI_TRIAL_TEAM_EMAIL", "anonymous-trial-user@example.com"
     )
     AI_TRIAL_REGION: str = os.getenv("AI_TRIAL_REGION", "eu-west-1")
+    # Age at which an unused trial key is reaped. Each trial provisions a
+    # LiteLLM key AND a Postgres database, and nothing else ever reclaims
+    # either, so without this they accumulate for the life of the region.
+    AI_TRIAL_RETENTION_DAYS: int = int(os.getenv("AI_TRIAL_RETENTION_DAYS", "30"))
+    # Cap per reaper run. Each deletion is an HTTP call plus a DROP DATABASE,
+    # so a first run against a large backlog is spread over several nights
+    # rather than held open for hours.
+    AI_TRIAL_REAP_BATCH_SIZE: int = int(os.getenv("AI_TRIAL_REAP_BATCH_SIZE", "500"))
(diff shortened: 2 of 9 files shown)


app/core/config.py
```diff
@@ -60,6 +60,14 @@ class Settings(BaseSettings):
         "AI_TRIAL_TEAM_EMAIL", "anonymous-trial-user@example.com"
     )
     AI_TRIAL_REGION: str = os.getenv("AI_TRIAL_REGION", "eu-west-1")
+    # Age at which an unused trial key is reaped. Each trial provisions a
+    # LiteLLM key AND a Postgres database, and nothing else ever reclaims
+    # either, so without this they accumulate for the life of the region.
+    AI_TRIAL_RETENTION_DAYS: int = int(os.getenv("AI_TRIAL_RETENTION_DAYS", "30"))
```
dan2k3k4:
```suggestion
    AI_TRIAL_RETENTION_DAYS: int = int(os.getenv("AI_TRIAL_RETENTION_DAYS", "90"))
```

When we talk about `AI Trial` for Drupal - we specifically mean the `anonymous trials` -> which get around `~$2` of budget. If the AI key reaches that budget then it can be cleaned up `~1-2 weeks` after reaching the budget.

If the *anonymous trial* has been around and not used the `~$2` budget, we originally did not plan any clean up date, because it is very likely people (non-Drupal folks) would start to set up a Drupal site but not touch on the AI keys until a month or two later, in that case, we suggest increasing `anonymous trials` to `90` days (if we have to limit it, because we're getting impacted a lot by trials being generated via spambots or such).

scripts/cleanup_trial_keys.py
```diff
@@ -0,0 +1,275 @@
...
+sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))
+
+from app.core.trial_cleanup import (  # noqa: E402
+    LiveTrialRegionError,
+    TrialCleanupSummary,
+    assess_selection,
+    delete_trial_key,
+    select_trial_keys,
+)
+from app.db.database import SessionLocal  # noqa: E402
```
dan2k3k4:
```suggestion
from app.db.database import SessionLocal
```

Not sure what the `  # noqa: E402` comments are for

.lagoon.yml
```diff
@@ -19,6 +19,13 @@ environments:
         schedule: "0 5 1 * *"
         command: python scripts/trigger_trial_recon_job.py
         service: backend
+      # Daily, not monthly: every trial provisions a LiteLLM key AND a Postgres
+      # database, and nothing else reclaims either. Monthly lets ~6,000 of each
+      # pile up between runs. Capped per run by AI_TRIAL_REAP_BATCH_SIZE.
+      - name: reap-trial-keys
```
dan2k3k4:
```suggestion
      - name: reap-trial-keys
```

I don't think we need comments in `.lagoon.yml`

.lagoon.yml
```diff
@@ -65,6 +72,13 @@ environments:
...
```
dan2k3k4:
```suggestion
      - name: reap-trial-keys
```

dan2k3k4 (approved):

PR #695, opened 2026-08-19
fix: release limited_resources counters when the owner goes away

dspachos:
Fixes amazeeio/moad#675.

## Problem

`limited_resources` counters only went up. `decrement_resource` existed but nothing called it.

- A team at its member cap could never take a replacement member: create → delete → create returned 402 "Team has reached their maximum user limit."
- Limit rows outlived the team or user that owned them (3,474 orphaned `owner_type=TEAM` rows on dev).
- `POST /users/{id}/add-to-team` took no seat at all, so a team could pass its cap through that door.

## Change

- `LimitService.delete_limits(owner_type, owner_id)` deletes every row of one owner and refuses the system defaults.
- `DELETE /users/{id}` and `POST /users/{id}/remove-from-team` give the member seat back. The delete path also removes the user's own rows.
- `DELETE /teams/{id}` removes the team's rows. Members survive the team, so their own rows stay.
- `POST /users/{id}/add-to-team` now runs the same check as creating a member outright, and gives the seat back if the join is rolled back.
- `check_team_user_limit` corrects the stored count against the real member count before it increments, the same pattern `check_key_limits` already uses. A path that forgets to release can no longer lock a team out for good.
- Migration `b2c3d4e5f6a7` deletes the rows whose team or user is already gone.

## Tests

`make backend-test`: 1435 passed, 2 skipped.

New: seat release on member delete and on remove-from-team, seat taken on add-to-team and 402 at capacity, team rows gone after team delete, `delete_limits` scoping and its system refusal, and the stale-count correction.

`test_check_team_user_limit_with_limit_service_at_capacity` now creates the 3 members it claims to have. It used to sit at a fabricated cap with zero members, which is the corrupt state this fix repairs.

## Reviewer notes

`add-to-team` can now answer 402, which it never did before. The admin frontend needs no change: `frontend/src/utils/api.ts:42` throws with the API `detail` and `use-teams.ts` shows it in an error toast, so an admin sees "Team has reached their maximum user limit." MOAD holds only the generated client for this endpoint, no call site.

### DIFF
--- a/app/api/teams.py
+++ b/app/api/teams.py
@@ -36,6 +36,7 @@ from app.db.models import (
     DBTeamRegion,
     DBUser,
 )
+from app.schemas.limits import OwnerType
 from app.schemas.models import (
     BudgetType,
     SalesProduct,
@@ -449,6 +450,10 @@ async def delete_team(team_id: int, db: Session = Depends(get_db)):
     db.query(DBBudgetAlertState).filter(DBBudgetAlertState.team_id == team_id).delete(
         synchronize_session=False
     )
+
+    # Limit rows point at the team by id with no foreign key, so they would stay
+    # behind for good and give a later team with the same id a used-up counter.
+    LimitService(db).delete_limits(OwnerType.TEAM, team_id, commit=False)
 
     # Delete the team
     db.delete(db_team)
--- a/app/api/teams.py
+++ b/app/api/teams.py
@@ -1053,7 +1053,13 @@ async def merge_teams(
                 except Exception as e:
                     logger.error(f"Failed to update LiteLLM key {key.id}: {str(e)}")
 
-        # Delete source team
+        # Delete source team. Its limit rows must go with it, same as in
+        # delete_team: owner_id has no foreign key, so they would outlive the
+        # team and give a later team with the same id a used-up counter.
+        # The target team takes the members without a cap check: the merge is
+        # a system-admin operation, and its member counter is re-derived from
+        # the real members on the next member addition.
+        LimitService(db).delete_limits(OwnerType.TEAM, source_team.id, commit=False)
         db.delete(source_team)
         db.commit()
 
--- a/app/api/users.py
+++ b/app/api/users.py
@@ -16,10 +16,10 @@ from app.core.litellm_user_sync import (
     sync_remove_user_from_team,
     sync_update_user_team_role,
 )
-from app.core.limit_service import LimitNotFoundError, LimitService
+from app.core.limit_service import LimitService
 from app.db.database import get_db
 from app.core.dependencies import get_limit_service
-from app.schemas.limits import OwnerType, ResourceType
+from app.schemas.limits import OwnerType
 from app.schemas.models import (
     User,
     AdminUserUpdate,
@@ -844,10 +844,6 @@ async def _create_user_in_db(user: UserCreate, db: Session) -> DBUser:
     # user-creation choke point (covers create_user, sign-in auto-provision, trial).
     assert_email_domain_allowed(db, user.email)
 
-    limit_service = get_limit_service(db)
-    if settings.ENABLE_LIMITS and user.team_id is not None:
-        limit_service.check_team_user_limit(user.team_id)
-
     # Validate role if provided
     if user.role and user.role not in UserRole.get_all_roles():
         raise HTTPException(
@@ -862,6 +858,12 @@ async def _create_user_in_db(user: UserCreate, db: Session) -> DBUser:
             status_code=status.HTTP_400_BAD_REQUEST,
             detail="system_admin role cannot be assigned via user creation.",
         )
+
+    # Take the seat only after every validation that can reject the request,
+    # so a rejected request does not hold a seat until the next recount.
+    limit_service = get_limit_service(db)
+    if settings.ENABLE_LIMITS and user.team_id is not None:
+        limit_service.check_team_user_limit(user.team_id)
 
     # Default to the lowest permissions for a user in a team
     if user.role is None and user.team_id is not None:
@@ -1047,18 +1049,6 @@ async def update_user(
     return db_user
 
 
-def release_team_user_seat(db: Session, team_id: int) -> None:
-    """
-    Give one member seat back to a team when a member leaves, is deleted, or never got in.
-
-    A team that never reached the limit service has no counter to release.
-    """
-    try:
-        LimitService(db).decrement_resource(OwnerType.TEAM, team_id, ResourceType.USER)
-    except LimitNotFoundError:
-        logger.info(f"Team {team_id} has no user limit to release")
-
-
 @router.post(
     "/{user_id}/add-to-team",
     response_model=User,
@@ -1115,8 +1105,6 @@ async def add_user_to_team(
         db.refresh(db_user)
     except Exception:
         db.rollback()
-        if settings.ENABLE_LIMITS:
-            release_team_user_seat(db, team_operation.team_id)
         raise
 
     try:
@@ -1126,8 +1114,6 @@ async def add_user_to_team(
             db_user.team_id = None
             db.commit()
             db.refresh(db_user)
-            if settings.ENABLE_LIMITS:
-                release_team_user_seat(db, team_operation.team_id)
         except Exception:
             db.rollback()
             logger.exception(
@@ -1217,11 +1203,8 @@ async def remove_user_from_team(
             )
         raise
 
-    # The counter went up when this member was created, so give the seat back.
-    # Done after the removal is final, so a failed removal keeps the seat taken.
-    if settings.ENABLE_LIMITS:
-        release_team_user_seat(db, previous_team_id)
-
+    # No counter update here: the member counter is re-derived from the real
+    # members on the next check, so the freed seat is picked up there.
     return db_user
 
 
@@ -1263,10 +1246,6 @@ async def delete_user(
     except Exception:
         db.rollback()
         raise
-
-    # The counter went up when this member was created, so give the seat back.
-    if settings.ENABLE_LIMITS and team_id is not None:
-        release_team_user_seat(db, team_id)
 
     try:
         await sync_delete_user_across_regions(db=db, db_user=db_user, team_id=team_id)
--- a/app/api/users.py
+++ b/app/api/users.py
@@ -1047,6 +1047,18 @@ async def update_user(
     return db_user
 
 
+def release_team_user_seat(db: Session, team_id: int) -> None:
+    """
+    Give one member seat back to a team when a member leaves, is deleted, or never got in.
+
+    A team that never reached the limit service has no counter to release.
+    """
+    try:
+        LimitService(db).decrement_resource(OwnerType.TEAM, team_id, ResourceType.USER)
+    except LimitNotFoundError:
+        logger.info(f"Team {team_id} has no user limit to release")
+
+
 @router.post(
     "/{user_id}/add-to-team",
     response_model=User,
@@ -1091,6 +1103,11 @@ async def add_user_to_team(
     if not db_team:
         raise HTTPException(status_code=404, detail="Team not found")
 
+    # This is the second door into a team, next to creating a member outright,
+    # so it has to take a seat as well. Without it a team can pass its cap.
+    if settings.ENABLE_LIMITS:
+        get_limit_service(db).check_team_user_limit(team_operation.team_id)
+
     # Add user to team
     db_user.team_id = team_operation.team_id
     try:
@@ -1098,6 +1115,8 @@ async def add_user_to_team(
         db.refresh(db_user)
     except Exception:
         db.rollback()
+        if settings.ENABLE_LIMITS:
+            release_team_user_seat(db, team_operation.team_id)
         raise
 
     try:
@@ -1107,6 +1126,8 @@ async def add_user_to_team(
             db_user.team_id = None
             db.commit()
             db.refresh(db_user)
+            if settings.ENABLE_LIMITS:
+                release_team_user_seat(db, team_operation.team_id)
         except Exception:
             db.rollback()
             logger.exception(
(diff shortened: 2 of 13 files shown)


app/core/limit_service.py
```diff
@@ -1005,6 +1045,8 @@ def check_team_user_limit(self, team_id: int) -> None:
 
         # First try the new service, and short circuit if it works
         try:
+            # Before trying to increment, validate and correct the current count if needed
+            self._validate_and_correct_team_user_count(team_id)
```
dan2k3k4: **Race: the recount erases concurrent in-flight reservations, allowing the cap to be overshot.**

`_create_user_in_db` commits the increment (limit_service.py:303) *before* the user row is committed (users.py:888), and there is no row locking (`with_for_update`) anywhere in the service. Team cap 5 with 4 committed members: request A increments to 5 (its user row not yet committed); request B's recount sees 4 committed members, resets the counter 5 → 4, increments to 5, and passes. Both commit → 6 members in a 5-cap team. Pre-PR, B would have received a 402.

This mirrors the pre-existing key-count pattern, but it is newly introduced for the USER resource here — and the reworked at-capacity test now pads real members so no test pins rejection when the counter exceeds committed membership (the reservation state). A `SELECT ... FOR UPDATE` on the limit row across correct+increment, or a single atomic conditional `UPDATE ... WHERE current_value < max_value`, closes it.

app/core/limit_service.py
```diff
@@ -1523,6 +1565,44 @@ def _validate_and_correct_service_key_count(self, team_id: int) -> None:
...
+        )
+
+        if not limit:
+            return  # No limit exists, will be created by the fallback logic
+
+        actual_count = self.db.execute(
+            select(func.count()).select_from(DBUser).where(DBUser.team_id == team_id)
+        ).scalar()
+
+        if limit.current_value != actual_count:
```
dan2k3k4: **The correction ignores `limited_by`, so MANUAL `current_value` overrides are silently reverted.**

`overwrite_limit` (app/api/limits.py:104-115) lets an admin set `current_value` with `LimitSource.MANUAL` — e.g. freezing a misbehaving team at cap. This function filters only on owner/resource and unconditionally overwrites `current_value` with the real member count on the next `check_team_user_limit`, undoing the manual freeze (or a manually granted grace) with only a log line. Consider skipping the correction (or requiring a larger drift threshold) when `limited_by == MANUAL`.

app/api/users.py
```diff
@@ -1046,6 +1047,18 @@ async def update_user(
...
 
+def release_team_user_seat(db: Session, team_id: int) -> None:
+    """
+    Give one member seat back to a team when a member leaves, is deleted, or never got in.
+
+    A team that never reached the limit service has no counter to release.
+    """
+    try:
+        LimitService(db).decrement_resource(OwnerType.TEAM, team_id, ResourceType.USER)
+    except LimitNotFoundError:
```
dan2k3k4: **Only `LimitNotFoundError` is caught; `decrement_resource`'s `ValueError` escapes after the delete already committed.**

The limits API has no cross-field validation tying `resource=USER` to CONTROL_PLANE/COUNT, so an admin can create a `(TEAM, id, USER)` row with `limit_type=DATA_PLANE` or a non-COUNT unit. `_verify_control_plane_count` then raises `ValueError`, which this helper does not shield. In `delete_user` the release runs after `db.commit()` and *before* `sync_delete_user_across_regions`, so the endpoint 500s with the user deleted locally but still alive in every LiteLLM region; every subsequent member deletion for that team 500s the same way. Catching `ValueError` here too (log and continue) keeps a bad limit row from breaking deletes.

app/api/teams.py
```diff
@@ -450,6 +451,10 @@ async def delete_team(team_id: int, db: Session = Depends(get_db)):
         synchronize_session=False
     )
 
+    # Limit rows point at the team by id with no foreign key, so they would stay
+    # behind for good and give a later team with the same id a used-up counter.
+    LimitService(db).delete_limits(OwnerType.TEAM, team_id, commit=False)
```
dan2k3k4: **`merge_teams` still has both halves of the bug this PR fixes.**

`app/api/teams.py:1006` moves every source member with `user.team_id = target_team.id` and no `check_team_user_limit` — the target absorbs members past its cap with a stale counter — and `db.delete(source_team)` at line 1051 has no `delete_limits`, so the source team's limit rows orphan exactly like `delete_team`'s used to. The one-time migration only cleans orphans that exist at deploy time; every merge afterwards creates new ones.

app/core/limit_service.py
```diff
@@ -334,6 +334,46 @@ def decrement_resource(
 
         return True
 
+    def delete_limits(
```
dan2k3k4: **Another deletion path still skips this: the trial-signup failure cleanup.**

`app/api/auth.py:1034-1044` creates a MANUAL USER BUDGET limit row before key creation (line ~1065) can fail, and the cleanup at ~1111-1114 does `db.delete(user); db.commit()` without `delete_limits(OwnerType.USER, ...)`. So new orphans keep appearing after the one-time migration — and these are MANUAL rows, which outrank DEFAULT/PRODUCT, so a future user allocated the same id is silently pinned to the trial budget cap. Routing that cleanup through `delete_limits` closes it.

app/api/users.py
```diff
@@ -1046,6 +1047,18 @@ async def update_user(
     return db_user
 
 
+def release_team_user_seat(db: Session, team_id: int) -> None:
```
dan2k3k4: **The primary seat-taking site, `_create_user_in_db`, never releases the seat on its own failure paths.**

`check_team_user_limit` at users.py:847-848 increments before the invalid-role / system_admin 400s a few lines below, and the compensating `db.delete(db_user)` when `sync_create_user_across_regions` fails (~line 895-899) also has no release. That is the same leak class this PR fixes in the other three endpoints, in the same file. The recount softens it to temporary drift, but if the explicit-release mechanism is kept, this path needs it most.

app/api/users.py
```diff
@@ -1090,13 +1103,20 @@ async def add_user_to_team(
     if not db_team:
         raise HTTPException(status_code=404, detail="Team not found")
 
+    # This is the second door into a team, next to creating a member outright,
+    # so it has to take a seat as well. Without it a team can pass its cap.
+    if settings.ENABLE_LIMITS:
+        get_limit_service(db).check_team_user_limit(team_operation.team_id)
```
dan2k3k4: **There is a third door: `_pin_user_to_key_team` (app/api/private_ai_keys.py:1466-1476) moves `current_user.team_id` across teams with no seat release on the old team and no check/increment on the new one.**

Same gap this hunk closes for add-to-team: the pin path admits a user into a team past its cap with a stale counter, and leaves the old team's counter one too high until the lazy correction happens to run. If the pin flow is intentionally exempt from the cap, a comment saying so would stop the next reader from re-flagging it.

app/api/users.py
```diff
@@ -1106,6 +1126,8 @@ async def add_user_to_team(
             db_user.team_id = None
             db.commit()
             db.refresh(db_user)
+            if settings.ENABLE_LIMITS:
+                release_team_user_seat(db, team_operation.team_id)
```
dan2k3k4: **The release is inside the revert `try`, so a release failure falsely logs "Failed to revert team assignment".**

By this point `db.commit()` already persisted the revert; if `release_team_user_seat` then raises (e.g. the `ValueError` case flagged on the helper), the `except` logs that the revert failed — it did not — and the seat stays leaked. Moving the release after the try/except (or into its own guard) keeps the log truthful and the failure domains separate.

app/api/users.py
```diff
@@ -1194,6 +1216,12 @@ async def remove_user_from_team(
                 db_user.id,
             )
         raise
+
+    # The counter went up when this member was created, so give the seat back.
+    # Done after the removal is final, so a failed removal keeps the seat taken.
+    if settings.ENABLE_LIMITS:
+        release_team_user_seat(db, previous_team_id)
```
dan2k3k4: **Double-failure path never reaches this release.**

If `sync_remove_user_from_team` raises AND the revert commit also fails, the outer `raise` fires before this line: the removal from the first commit stands (user permanently off the team) but the seat is never released. Rare, but it is exactly the "path that forgets to release" this PR describes — worth a note that the recount is the backstop, or releasing in a `finally`-style guard keyed on whether the removal is still committed.

app/api/users.py
```diff
@@ -1046,6 +1047,18 @@ async def update_user(
...
 
 
+def release_team_user_seat(db: Session, team_id: int) -> None:
+    """
+    Give one member seat back to a team when a member leaves, is deleted, or never got in.
+
+    A team that never reached the limit service has no counter to release.
+    """
+    try:
+        LimitService(db).decrement_resource(OwnerType.TEAM, team_id, ResourceType.USER)
```
dan2k3k4: **Design tension: this PR ships two mechanisms maintaining the same number — pick one.**

The recount in `check_team_user_limit` re-derives `current_value` from `COUNT(*)` before every increment, which makes all four explicit `release_team_user_seat` call sites (including both rollback-compensation blocks) redundant — keys already live recount-only, with zero `decrement_resource` callers anywhere in app/api. Conversely, if explicit decrements are the model, the every-request recount (an unindexed `COUNT(*)` over `users` — `users.team_id` has no index — plus an extra SELECT and possible extra commit on each member creation) is redundant on the happy path and could run only when the increment fails.

The unconditional decrement half also misbehaves for pool-gated teams: `check_team_user_limit` returns at the gate before both the recount and the increment, but deletes still decrement, so a gated team's counter decays toward 0 until the first post-gating check corrects it (reported usage is wrong meanwhile). Recount-only — matching the existing key pattern — deletes ~30 lines and this whole class of asymmetry.

dspachos: Done in 7127a39a: recount-only, matching the key counters. The counter is re-derived from COUNT(*) inside the check, in one locked transaction. The release helper and all four call sites are gone, and the pool-gate decay with them. Also added an index on users.team_id for the per-check count (new migration).

app/core/limit_service.py
dspachos: Fixed in 7127a39a: the correction and the increment now run in one transaction with the limit row locked (SELECT FOR UPDATE), so two checks cannot lose each other's update. The residual race stays: the count cannot see a member row whose transaction has not committed, so two concurrent additions can still pass at the last seat — same as the existing key counters. Closing that would need the seat and the member row in one transaction; the docstring states the limitation.

app/core/limit_service.py
dspachos: Valid concern, but with recount-only (see the pick-one thread) a MANUAL skip would be unsafe: nothing decrements the counter any more, so a skipped MANUAL team would creep to its cap and stay locked out — the exact bug this PR fixes. So the recount applies to MANUAL rows too. The docstring in 7127a39a now says a hand-set current_value does not survive the next check; freeze a team by lowering max_value instead.

app/api/users.py
dspachos: Gone in 7127a39a: the release helper and all decrement call sites are removed (recount-only). delete_user now only calls delete_limits, which raises only for SYSTEM owners.

app/api/teams.py
dspachos: Fixed in 7127a39a: merge_teams now deletes the source team's limit rows before deleting the team, same as delete_team. The target absorbs members without a cap check on purpose — the merge is a system-admin operation — and its counter is re-derived from the real members on the next member addition. Both points are documented at the site, and a test covers the row cleanup.

app/core/limit_service.py
dspachos: Fixed in 7127a39a: the trial cleanup now calls delete_limits(USER, ...) in the same commit as the user delete, so the MANUAL budget row dies with the user. The cleanup test asserts the call.

app/api/users.py
dspachos: Fixed in 7127a39a: the limit check moved after the role validations, so a rejected request never takes a seat. The sync-failure path needs no compensation any more — the recount re-derives the counter on the next check.

app/api/users.py
dspachos: Intentional exemption, now documented in the docstring (7127a39a): moad already created the key under the target team, and rejecting the pin would strand a live key the user can never see. Both teams' counters are re-derived on their next member addition.

app/api/users.py
dspachos: Gone in 7127a39a together with the release mechanism — the revert block no longer contains a release, so the log stays truthful.

app/api/users.py
dspachos: Gone in 7127a39a together with the release mechanism. The recount is the backstop for this path; a comment at the site says so.

app/migrations/versions/20260819_100000_b2c3d4e5f6a7_cleanup_orphaned_limited_resources.py
```diff
@@ -0,0 +1,46 @@
...
+Create Date: 2026-08-19 10:00:00.000000+00:00
+
+"""
+
+from typing import Union
+
+from alembic import op
+
+# revision identifiers, used by Alembic (read via module reflection).
+revision: str = "b2c3d4e5f6a7"
```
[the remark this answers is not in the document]
dspachos: False positive: Alembic reads revision and down_revision from the module via reflection.

app/migrations/versions/20260819_100000_b2c3d4e5f6a7_cleanup_orphaned_limited_resources.py
```diff
@@ -0,0 +1,46 @@
...
+down_revision: Union[str, None] = "f9a0b1c2d3e4"
```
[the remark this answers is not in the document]
dspachos: False positive: Alembic reads revision and down_revision from the module via reflection.

dan2k3k4 (approved):

PR #697, opened 2026-08-19
Derive team admin_email from the authenticated caller

dan2k3k4:
`admin_email` on a team is now derived from the authenticated caller instead of taken from the request body.

### What
- `register_team`: a non-admin caller becomes the admin of the team they create; the body value is ignored. System admins may still name another admin — the admin UI and support flows need that.
- `register_team`: `current_user` is now `Optional`. The two internal (non-HTTP) callers — sign-in auto-provision and anonymous trial team creation — pass `current_user=None` and keep setting `admin_email` themselves.
- `register_team`: the disposable-domain check moved after the override, so it validates the address that is actually stored instead of a discarded body value.
- `update_team`: `admin_email` joins the admin-only field set, compared case-insensitively so a GET-then-PUT round-trip that re-sends the current address is still a no-op. A changed value also goes through the disposable-domain check and the same case-insensitive uniqueness check `register_team` has — the DB unique index is case-sensitive, so a case variant would otherwise be stored silently, and an exact duplicate surfaced as an unhandled IntegrityError.

Consistency matters on the update path too: sign-in re-attaches a first-time user to the team that already carries their address (`app/api/auth.py`), so `admin_email` decides team membership and is not a free-text field.

### Out of scope
- Restricting `POST /teams` to platform admins: self-service team creation is intended.
- Rate limiting team creation.

### Testing
`make backend-test` — 1450 passed, 2 skipped. 9 new tests in `tests/test_teams.py`. `ruff check` and `ruff format` clean on the changed files. Rebased onto current `dev`.

Closes amazeeio/moad#721

### DIFF
--- a/app/api/auth.py
+++ b/app/api/auth.py
@@ -495,7 +495,7 @@ async def sign_in(
                 budget_type=BudgetType.PERIODIC,
                 region_id=sign_up_region.id,
             )
-            team = await register_team(team_data, db)
+            team = await register_team(team_data, db, current_user=None)
 
         user_data = UserCreate(
             email=sign_in_username,
@@ -991,7 +991,7 @@ async def generate_trial_access(
                 is_active=True,
                 budget_type=BudgetType.PERIODIC,
             )
-            team = await register_team(team_data, db)
+            team = await register_team(team_data, db, current_user=None)
             # Ensure team has limit set
             team_limit = limit_service.set_limit(
                 owner_type=OwnerType.TEAM,
--- a/app/api/teams.py
+++ b/app/api/teams.py
@@ -412,11 +412,28 @@ async def update_team(
                 ),
             )
 
-    # Same blocklist gate as team creation, but only when the value changes.
+    # Same gates as team creation, but only when the value changes.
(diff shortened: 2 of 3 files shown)


dspachos:
The intent is right, but this breaks two internal call paths:

1. `app/api/teams.py:126` — `register_team` is also called directly (not through FastAPI) at `app/api/auth.py:498` and `app/api/auth.py:994` with only `(team_data, db)`. In those calls `current_user` stays the `Depends(...)` default object, so `current_user.is_admin` raises `AttributeError`. Sign-in auto-provision and trial-team creation will return 500 — the exact paths the PR description says it preserves. We should pass an explicit user (or `None` with a guard) in the direct calls.
2. `app/api/teams.py:123` — `assert_email_domain_allowed` runs on the request-body `admin_email` before the override, so the value that is actually stored (`current_user.email`) is never checked, and a discarded request value can still reject the call. The check should run on the effective email after the override.
3. Nit: all `register_team` tests use `admin_token` (`tests/test_teams.py:43`); no test covers the new non-admin override path, so a regression here (including point 1) is not caught by the team tests.

dspachos (approved):

PR #698, opened 2026-08-19
Read validate-jwt token from the Authorization header only

dan2k3k4:
`GET /auth/validate-jwt` reads the token only from the `Authorization: Bearer` header. The `token` query parameter is removed, and the docstring is updated.

### Why
Query strings are recorded by proxies, browser history and monitoring, so they are the wrong place for a credential. The header is the form every caller already uses.

### What
- The endpoint reads the header only.
- Tests: the four `*_query_param` tests in `tests/test_jwt.py` are removed — each had an identical `*_auth_header` twin, so no coverage is lost — and replaced by `test_validate_jwt_rejects_query_param_token`, which asserts a valid token in the query string yields 401, issues no new token and sends no recovery email.
- The cookie-expiration tests (`tests/test_auth.py`) and the revocation / missing-`jti` / expired-recovery tests (`tests/test_logout_revocation.py`) were converted to header form, so they keep exercising their original behaviour rather than passing because the query token is now ignored.

### Breaking change
Any caller still passing `?token=` gets a 401 and must move to the header. No in-house caller is affected: the emailed recovery link points at the frontend `/upgrade?token=...` page, and the frontend calls this endpoint with a header (`frontend/src/utils/api.ts`).

Follow-up in another repo: moad's generated OpenAPI client still describes this endpoint with a `?token=` query parameter (`src/external/generated/amazeeai.live.ts`, `src/server/external/generated/amazeeai.live.ts`). No hand-written moad code calls it, so nothing breaks today, but the client should be regenerated.

### Out of scope
A dedicated refresh-token mechanism with an absolute lifetime.

### Testing
`make backend-test` — 1438 passed, 2 skipped. `ruff check` clean. Rebased onto current `dev`.

Closes amazeeio/moad#723

### DIFF
--- a/app/api/auth.py
+++ b/app/api/auth.py
@@ -748,15 +748,12 @@ async def delete_token(
 async def validate_jwt(
     request: Request,
     response: Response,
-    token: Optional[str] = None,
     db: Session = Depends(get_db),
 ):
     """
     Validate a JWT token and either refresh it or send a new validation URL.
 
-    The token can be provided either:
-    - As a query parameter: ?token=your_token
-    - In the Authorization header: Bearer your_token
+    The token must be provided in the Authorization header: Bearer your_token
 
     Returns:
     - If token is valid: A new access token with cookies set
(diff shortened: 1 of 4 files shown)


dspachos:
The endpoint change itself is correct, and the frontend already sends the Bearer header (the emailed recovery URL points to the frontend `/upgrade?token=...` page, which calls the endpoint with a header). But the tests that use `?token=` were not updated, so CI will fail:

1. `tests/test_jwt.py:54` (`test_validate_jwt_valid_token_query_param`) — expects 200, now gets 401.
2. `tests/test_jwt.py:104` — the expired-token query-param test expects the "Token expired. A new validation URL has been sent..." detail and one SES call; the endpoint now returns "Could not validate credentials" with no email. Both assertions fail.
3. `tests/test_auth.py:929` and `:947` — both cookie-expiration tests use `?token=` and expect 200; they now get 401.
4. `tests/test_logout_revocation.py:135` and `:197` — line 135 asserts 200 for a query-param call and line 197 expects the "validation URL has been sent" detail; both fail.
5. `tests/test_logout_revocation.py:156` and `:178` — these still pass, but only because the query token is now ignored; they no longer exercise the revocation and missing-`jti` checks. Please convert them to header form, or we silently lose that coverage.

Also keep in mind that removing the query parameter is a breaking API change for any external caller, as the description states.

dspachos (approved):

PR #699, opened 2026-08-19
Set SameSite=Lax on the access_token cookie

dan2k3k4:
The `access_token` cookie is issued with `SameSite=Lax` instead of `None`. `Secure` is unchanged.

### What
- Both cookie sites change together: `create_and_set_access_token` (`response.set_cookie`) and `logout` (`response.delete_cookie`), and the logout call now also sets `HttpOnly` so the two are identical. Removal is keyed on name, path and domain, so that last part changes no behaviour; it stops the two call sites drifting apart unnoticed.
- Two tests in `tests/test_auth.py` pin the attributes: the login cookie carries `Secure`, `HttpOnly` and `SameSite=Lax`, and the logout cookie repeats them. Nothing asserted `SameSite` before, so a change back would have passed CI.

### Deploy check
All deployed frontends are same-site with their backend, so cookie auth is unaffected:

| Env | Frontend | Backend | Cookie domain |
|---|---|---|---|
| PROD | `dashboard.amazee.ai`, `amazee.ai` | `api.amazee.ai` | `amazee.ai` |
| DEV | `frontend.dev.amazeeai.us2.amazee.io` | `backend.dev.amazeeai.us2.amazee.io` | derived from `LAGOON_ROUTES` |

moad (`my.amazee.io`) and the Drupal module authenticate with Bearer headers, not this cookie. Note that CORS still accepts `FRONTEND_ROUTE` origins on any domain, so a future cross-site frontend deployment would need a different auth transport.

### Out of scope
The moad `sessionId` cookie.

### Testing
`make backend-test` — 1443 passed, 2 skipped. `ruff check` clean. Independent review found no issues. Rebased onto current `dev`.

Closes amazeeio/moad#727

### DIFF
--- a/app/api/auth.py
+++ b/app/api/auth.py
@@ -274,6 +274,7 @@ async def logout(
     delete_settings = {
         "key": "access_token",
         "path": "/",
+        "httponly": True,
         "secure": True,
         "samesite": "lax",
     }
--- a/app/api/auth.py
+++ b/app/api/auth.py
@@ -183,7 +183,7 @@ def create_and_set_access_token(
         "httponly": True,
         "max_age": cookie_expiration,
         "expires": cookie_expiration,
-        "samesite": "none",
+        "samesite": "lax",
         "secure": True,
         "path": "/",
     }
@@ -275,7 +275,7 @@ async def logout(
         "key": "access_token",
         "path": "/",
         "secure": True,
-        "samesite": "none",
+        "samesite": "lax",
     }
 
     # Only set domain if we got one from LAGOON_ROUTES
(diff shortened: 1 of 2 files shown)


dspachos:
I checked the consumers: `set_cookie` (`app/api/auth.py:186`) and `delete_cookie` (`app/api/auth.py:278`) are the only two `samesite` sites and they match, so logout still clears the cookie. All in-repo consumers are same-site (sibling subdomains under the shared cookie domain) or use Bearer headers (Drupal, moad). Email sign-in links point to the frontend, so no cross-site top-level POST to the backend depends on the cookie. Two points:

1. CORS (`app/main.py:137-147`) accepts `FRONTEND_ROUTE` origins on any domain, so the code supports frontends that are cross-site to the backend. With `Lax`, such a deployment loses cookie auth: login succeeds but every later request is unauthenticated. The description already flags this — let's verify deploy config before merge.
2. `tests/test_auth.py:807` — tests assert `Max-Age` on `Set-Cookie` but nothing asserts `SameSite`, so a revert to `None` would pass CI.

dspachos (approved):

PR #700, opened 2026-08-19
Scope the OpenAPI schema to the caller's access level

dan2k3k4:
`/openapi.json` is now served per caller: anonymous callers get the operations that need no authentication, authenticated users additionally get the authenticated ones, and system administrators get the full document. The schema and the Swagger UI at `/` stay reachable without auth — they are the API's docs.

This is visibility only. Every endpoint still enforces its own authorization; nothing about access control changes.

### How
- Each operation's tier comes from its own route dependencies, so the mapping stays in sync as endpoints are added: `get_role_min_system_admin` and `require_system_admin` mean admin, other auth dependencies mean user, none means public.
- The map is keyed on `operationId`, the one identifier shared by the route object and the generated document. Keying on `(method, path)` silently misses routes whose path carries a converter (`/{id:path}`).
- In this FastAPI version, routes added with `include_router` are not flattened into `app.routes`; the app holds a router wrapper that resolves its operations on demand. `_iter_operations` handles both shapes. Because that is internal FastAPI API, a test asserts that every operation in the document is present in the map, so a FastAPI upgrade that changes the shape fails CI instead of quietly emptying the public docs.
- An operation missing from the map is treated as admin-only, so the failure mode is a smaller document rather than a wider one. Same for an RBAC dependency used as an instance (`Depends(require_system_admin())`), which carries no `__name__` to match on: those are classified by the roles they allow.
- `/metrics` is excluded from the document. It carries no route dependency, and it is only registered when `ENABLE_METRICS` is on — the deployed setting, but not the test one — so it would otherwise have been listed publicly while `AuthMiddleware` answers it with a 404 without the Prometheus key.
- `check_sales_or_higher` counts as admin: it guards a staff-only bulk endpoint.
- The caller is resolved with the normal auth dependency, so JWTs, API tokens and the local bearer token all count as authenticated. Anything that does not resolve is anonymous.
- Request and response models reachable only from hidden operations are pruned transitively, along with tag descriptions that no longer label a visible operation. `deepcopy` protects the cached full document.

`CacheControlMiddleware` already sets `no-store` on this path, so per-caller documents are not proxy-cached.

### Out of scope
The same tiered idea for moad's GraphQL introspection, and the LiteLLM instance's own docs.

### Testing
`make backend-test` — 1459 passed, 2 skipped. 17 new tests in `tests/test_openapi_scoping.py`. `ruff check` clean. Rebased onto current `dev`.

Closes amazeeio/moad#724

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -30,7 +30,7 @@ from app.middleware.audit import AuditLogMiddleware
 from app.middleware.auth import AuthMiddleware
 from app.middleware.caching import CacheControlMiddleware
 from app.middleware.prometheus import PrometheusMiddleware
-from fastapi import FastAPI
+from fastapi import Depends, FastAPI, Request
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.middleware.trustedhost import TrustedHostMiddleware
 from fastapi.openapi.docs import (
@@ -38,6 +38,14 @@ from fastapi.openapi.docs import (
     get_swagger_ui_oauth2_redirect_html,
 )
 from fastapi.openapi.utils import get_openapi
+from fastapi.responses import JSONResponse
+from fastapi.routing import APIRoute
+from copy import deepcopy
+from jose import jwt
+from sqlalchemy.orm import Session
+from app.db.database import get_db
+from app.api.users import get_user_by_email
+from app.core.security import assert_token_not_revoked
 from prometheus_fastapi_instrumentator import Instrumentator, metrics
 from starlette.middleware.base import BaseHTTPMiddleware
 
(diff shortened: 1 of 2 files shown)


dspachos:
The approach is sound (the schema stays in `PUBLIC_PATHS`, `no-store` prevents proxy caching, `deepcopy` protects the cached full schema, and the tier map builds after all routers). But two gaps make the filtering incorrect:

1. The `_OPENAPI_TIERS` map keys on `route.path`, but OpenAPI path keys use `route.path_format`, which strips converters. The admin-only route `GET /admin/models/{model_id_or_id:path}` (`app/api/admin_models.py:233`) misses the lookup and falls back to the "user" tier, so every logged-in non-admin sees this admin operation. Build the map from `route.path_format`.
2. `_caller_tier` only decodes JWTs. The API also accepts DBAPIToken bearer tokens and the local bearer token (`app/core/security.py:206-219`, `:190-204`), so a user or admin calling with an API token is classified "public" and gets the anonymous schema.
3. An operation missing from the map defaults to "user", which fails open (see point 1). Default to "admin" to fail closed.
4. `components.schemas` and `tags` are not filtered, so anonymous callers still see all request/response models of hidden admin endpoints.
5. Nit: no new tests; only the old anonymous-200 test exists (`tests/test_security.py:215`). Add tests that an admin path is hidden from anonymous and user callers.

dspachos (approved):

